PR #20, opened 2012-10-27
Making iso packagers support wipe/protect for PCI compliance

chhil:
Diff before commit http://screencast.com/t/S07t37fZx
Modified the jpos framework to support ISO message logging to enbale pci
compliant logging via configuration of the packager itself.
An additional optional attribute called "display"is added in the
packager dtd which can take values 0 = default, 1 = protect, 2=wipe

This removes the dependency of adding an additional ProtectedLogListener
(logs messages received and sent) and configuring a ProtectDebugLog (to
protect the iso messages in the context).
These are kind of duplicate configuration to support PCI compliant
logging that can lead to errors.
The older way did not give us control to log unique field (e.g. if field
61 for incoming request from Interface 1 needed to be protected but
field 61 for incoming request from Interface 2 did not need to be
protected. Thus we did not have control of the individual fine grained
control fields.

This change is at the packager level hence we can protect different
fields in different interfaces.

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/Display.java
+++ b/jpos/src/main/java/org/jpos/iso/Display.java
@@ -1,0 +1,8 @@
+package org.jpos.iso;
+
+public class Display {
+
+    public static final int WIPE    = 2;
+    public static final int PROTECT = 1;
+    public static final int NOP     = 0;
+}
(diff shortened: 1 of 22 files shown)


yinheli:
heavy implication.

PR #47, opened 2014-03-26
Replace TransactionManager thread management with thread pool

demsey:
Please read commits descriptions
It is for review and discuss here.

ISSUES:
 1 A bit unclear configuration of queuing methods (backward capability)
 2 Input space has to be LocalSpace to register listener
 3 Removed 'hack' to stop thread with Boolean.FALSE
 4 Removed checking pool size and resheduling task in future


ar:
Thank you for your great contribution @demsey as always, but I believe the next big thread related change for the TM will be to use fibers (virtual threads) right away. TM is working great, really stable on high load systems, supporting over 4000 TPS sustained load. I don't think we want to rock the boat with current implementation.

demsey:
Of course:). There's no problem.

PR #71, opened 2014-09-27
Standalone BItMap - for LIS5, but perhaps portable?

marksalter:
I need to consider the potential moveable nature of this BitMap field and form a method of keeping track of it/them in the ISOBasePacker processing, since field 65 (for LIS) is currently hardcoded.


marksalter:
Trimmed my fork, had forgotten about this PR, but I am sureit doesn't really matter :-)

PR #78, opened 2015-01-01
Improve LogEvent timestamp logging

demsey:
In log events tiestamps look like  at="Fri Dec 19 14:47:44 CET 2014.86"
After the change it will look like at="Fri Dec 19 14:47:44.086 CET 2014"

### DIFF
--- a/jpos/src/main/java/org/jpos/util/LogEvent.java
+++ b/jpos/src/main/java/org/jpos/util/LogEvent.java
@@ -26,6 +26,8 @@ import java.io.ByteArrayOutputStream;
 import java.io.IOException;
 import java.io.PrintStream;
 import java.sql.SQLException;
+import java.text.DateFormat;
+import java.text.SimpleDateFormat;
 import java.util.*;
 
 /**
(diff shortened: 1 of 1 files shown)


demsey:
Yes Date.toString() uses Locale.US, I checked it before ;)

I would be very careful when you do df as a static. SimplaDateFormat is not thread safe. (I have a bad experience when I did once df static - in production). It seems to me that LogEvent.dump() **can be called by different threads, and this will lead to disaster.**
I assumed that, multiple creation is a little slower but it works properly. I do not know how much slower.

Other ways to to deal with this problem: http://stackoverflow.com/questions/10411944/java-text-simpledateformat-not-thread-safe Will review what are the possibilities.

ar:
Oh, good point. Will revert my change for now.

demsey:
Maybe jast use FastDateFormat (commons-lang) instead of SimpleDateFormat?

ar:
I prefer not to add a new dependency just for one class ...

demsey:
I think so too;) I thought that the commons-lang is already in the jpos dependencies, but I was wrong...
I made some tests and it seems to me that newSimpleDateFormat is not so bad in this situation.
Thus it is slower but not 100 or even 10 times but 50%
```java
df.format(d);
```
This is "only" about 50% slower as above
```java
DateFormat df = new SimpleDateFormat(SOME_PATTERN);
df.format(d);
```
Here are the tests which that indicate https://github.com/demsey/jPOS/commits/logging

ar:
Thank you for the follow-up comments and tests Robert!

PR #82, opened 2015-01-23
Add ARQC veryfication and ARPC generation support in JCESecurityModule

demsey:
JUnit tests have been prepared with pay special attention on the basis of the results of real HSM.


vishnupillai:
Nice work! This will be very useful for people starting to implement EMV Cards

ar:
Awesome @demsey

PR #86, opened 2015-04-27
SMAdapter.translateKeyScheme and JCESecurityModule implementation

demsey:
HSM enables to convert double and triple length DES key encrypted
using X9.17 methods to more secure key using the variant method.
In theory is possible do it using export a key under a ZMK and
import it again under LMK with more secure encryption scheme,
but very often exporting and importing X9.17 keys is locked.
Translate Key Scheme to variant methods is enabled by default.

Also others cleaning patches;)


yinheli:
good job

PR #88, opened 2015-05-23
Fix thread allocation error with race conditions and improve thread naming

vishnupillai:
Bug scenario:
1. An ISO server receiving 2 connections on the same port
2. When the ISO Server is deployed, there are client channels trying to connect every 1 second. Race condition issue possible when both connections come in around the same time.
3. The ISO server receives both connections and creates the Session object, but the thread pool creates only 1 thread

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/iso/QServer.java
+++ b/jpos/src/main/java/org/jpos/q2/iso/QServer.java
@@ -102,7 +102,7 @@ public class QServer
         }
 
         ThreadPool pool = null;
-        pool = new ThreadPool (minSessions ,maxSessions);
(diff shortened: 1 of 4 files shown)


ar:
Agree with @tony4d, this is a +1000 awesome finding.

jatinsalla2003:
@vishnupillai is a rock star :+1:  1000000++++

hadixoy2:
Very good report thank you vishnupillai

hadixoy2:
I have other problem maybe related this
when max pool reached, some other ports in QServer that deployed, revive message but not process ( not apply filetrs and process)  i am trying to make Bug scenario

vishnupillai:
@hadixoy2 Thanks. Looking forward to your findings.

yinheli:
+1000

PR #100, opened 2016-02-24
Update ISO-4217 currencies

demsey:
* add missing: ZMW, TMM, SSP
* for other update replecement information

btw
I think the time has come to make use of the base of currencies provided by java `java.util.Currency`
Since java 7 are `getNumericCode()`, `getDisplayName()`.


ar:
> I think the time has come to make use of the base of currencies provided by java java.util.Currency
Since java 7 are getNumericCode(), getDisplayName().

I agree. Will gauge the impact.

PR #101, opened 2016-02-24
ISOCurrency now uses currencies from jvm

vsalaman:
Follows suggestion provided at [https://github.com/jpos/jPOS/pull/100]()


demsey:
Two questions (I'm not shure).

- Does jvm provide historical codes?
- and does this historical codes are stil required?

vsalaman:
No idea :)

ar:
With the new patch, many currencies were `numDecimals` == -1

```
XTS,963,-1
XSU,994,-1
XAG,961,-1
XUA,965,-1
XFO,000,-1
XAU,959,-1
XBA,955,-1
XDR,960,-1
XXX,999,-1
XFU,000,-1
XPD,964,-1
XPT,962,-1
XBD,958,-1
XBB,956,-1
XBC,957,-1
```

not sure what a -1 mean here. Not a big deal as these pseudo currencies were not available in `ISOCurrencies.properties` but perhaps we can protect against a -1 in the rest of the code (or just store it as zero).

Here is the [Javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Currency.html#getDefaultFractionDigits--)

"In the case of pseudo-currencies, such as IMF Special Drawing Rights, -1 is returned."

PR #106, opened 2016-04-01
header log bug fix

name327:


ar:
Thank you for your PR, but there's an `override-header` property that does what you need and has the option to not override it if you want. No need for this patch, unless I'm missing something, in which case, I'm all ears.

PR #107, opened 2016-04-07
There is no reason to declare never thrown ISOException

demsey:
In `ISOMsg` methods `set` and `unset` declare never thrown `ISOException`.
This leads to the introduction of many lines of boilerplate code at
calling that methods.
An example:
```java
  ISOMsg m = new ISOMsg("1120");
  try {
      m.set(39, "000");
  } catch (ISOException iggore) {} //NOPMD: never thrown
```
This change removes that unnececary declarations. **As a side effect it
can cause not compiling code.** It may be necessary to remove unnecesary
handling of `ISOException` (As a result simplifying that calls). Examples in
`CardHolder.java`, `ChannelInfoFilter.java` and `ISOMsgPanel.java`

### DIFF
--- a/jpos/src/main/java/org/jpos/core/CardHolder.java
+++ b/jpos/src/main/java/org/jpos/core/CardHolder.java
@@ -108,24 +108,20 @@ public class CardHolder implements Cloneable, Serializable, Loggeable {
         throws InvalidCardException
     {
         super();
-        try {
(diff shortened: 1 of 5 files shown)


ar:
Fair enough. May require some user code change, but I think it's a good change. Will merge right away.

PR #108, opened 2016-06-10
Fixed ISOUtil.byte2BitSet(...) for certain cases when 24 bytes are requested

barspi:
also some other related fixes and javadocs, indentation, etc


ar:
Awesome BBB!

PR #112, opened 2016-08-29
Converted into functional interface 

alcarraz:
Converted into functional interface by providing a default implementation of expired() method.


alcarraz:
Will add some example to docs ASAP

PR #115, opened 2016-09-11
Prevention against locking of deploy directory scanning

demsey:
Synchronizing on `this` may lead to locking of deploy directory scanning.
This happens when custom `QBean` implementation also use synchronization on `this`. Then `Q2.checkModified()` hangs when that custom `QBean` is locked on `this`.
In order to prevent such a case, `QBeanSupport` should use some special object for synchronization.

To solve this, I have used a new object:
`final Object modifyLock = new Object();`
I do not understand why the method  `getPersist()` was synchronized but `setPersist()` was not.
I think that `setPersist()` sould not be synchronized at all, but I'm not sure.
It is proposal only. So please check it.


ar:
Good catch. Thank you!

ar:
Sorry @demsey this pull request was left behind - will merge it right away.

PR #119, opened 2017-01-11
Serialization of a Configurable object enabled

TheBerserk:
There is need to serialize Configurable and XMLConfigurable objects.
In example to persist them.

### DIFF
--- a/jpos/src/main/java/org/jpos/core/Configurable.java
+++ b/jpos/src/main/java/org/jpos/core/Configurable.java
@@ -19,13 +19,15 @@
 package org.jpos.core;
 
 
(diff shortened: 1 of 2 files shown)


ar:
Can you expand on the use case for this? Many many objects implement `Configurable` in jPOS, having `Configurable` extend `Serializable` has a big impact with many side effects. Why don't you just implement Serializable on the objects you do want to serialize?

TheBerserk:
Wow, it's by mistake, should be `Configuration` instead of `Configurable`.
Don't know how i did it. I'll fix it in other branch.
Could You close this request please ?

PR #124, opened 2017-03-21
Avoid possible NPE when throwing ISOExceptions

bjansen:
While testing what would happen if a field was present in the bitmap but not configured in the "isopackager" XML file, I noticed that the original exception was lost because of an NPE. All I got in the end was a confusing message `null` instead of the more helpful `field packager '8' is null`. This should now be fixed :)


ar:
Awesome - thank you! Will merge it right away.

PR #142, opened 2017-10-24
[URGENT] Fix bad recursion in TransactionManager at commit #da2f535bb

barspi:


ar:
Thanks for the fix. I've already applied a similar fix before I saw the PR, sorry.

PR #156, opened 2018-03-20
Echo on Connect

ibrahimlawal:
Add setting which when true will prompt the Channel to send an echo right after connecting

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/BaseChannel.java
+++ b/jpos/src/main/java/org/jpos/iso/BaseChannel.java
@@ -31,10 +31,14 @@ import org.jpos.util.NameRegistrar;
 import javax.net.ssl.SSLSocket;
 import java.io.*;
 import java.net.*;
+import java.text.SimpleDateFormat;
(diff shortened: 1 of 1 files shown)


ar:
Thank you for the PR, but BaseChannel has a `postConnectHook()` that a channel implementation can use to send a logon/echo if required. I don't think all this hardcoded 0800 belongs to this already-polluted base class.

ibrahimlawal:
@ar Lol, very true. Concluded the PR before I checked it over. I even have code there that forced my TimeZone! 

I tried using the `postConnectHook()` to send the echo but for some reason, attempting to send before having set it as `usable` or calling `setChanged` was causing errors. Seeing as many implementations end up needing a signon, It felt flag-worthy :)

ar:
If you feel `postConnectHook` needs review just let us know, perhaps you can describe the problem you had in an issue and we can work together in a fix.

ibrahimlawal:
@ar true. Definitely worth it to check again. Will do so soon.

PR #157, opened 2018-03-20
Add JSONChannel and JSONPackager

ibrahimlawal:
Sketchy JSON Implementation. Expects non-prettified JSON (single line for all messages) of which these are samples:

**Flat**
```json
{"mti":"0800","data":{"3":"401010","7":"0320171042","11":"447767","12":"171042","13":"0320","70":"001"}}
```

**With subfields**
```json
{"mti":"0200","data":{"3":"401010","7":"0320171042","11":"447767","12":"171042","13":"0320","70":"001","127":{"data":{"3":"401010","7":"0320171042","11":"447767","12":"171042","13":"0320","70":"001"}}}}
```

### DIFF
--- a/jpos/build.gradle
+++ b/jpos/build.gradle
@@ -14,6 +14,7 @@ dependencies {
     compile libraries.slf4j_api
     compile libraries.javassist
     compile libraries.hdrhistogram
+    compile libraries.json
 
     testCompile libraries.hamcrest
     testCompile libraries.fest_assert
--- a/jpos/libraries.gradle
+++ b/jpos/libraries.gradle
@@ -1,6 +1,7 @@
 ext {
     libraries = [
         jdom: 'org.jdom:jdom2:2.0.6',
+        json: 'org.json:json:20180130',
         jdbm: 'jdbm:jdbm:1.0',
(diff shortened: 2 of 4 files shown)


ar:
Have you take a look at our [JSONPackager on hold](https://github.com/jpos/jPOS/blob/master/jpos/src/main/java/org/jpos/iso/packager/JSONPackager.hold) ? Although it's on hold here in jPOS due to some dependency conflicts, we have some deployments of it and we'd like the JSON to be compatible with it at the wire level. Do you think your implementation actually is?

ibrahimlawal:
@ar I checked the `JSONPackager on hold` file and attempted to  include it, but seeing no sample JSON to work with, I ended up attempting mine. The one thing I have not been able to test is sending Binary fields. I have not had any experience with those. What would be a good way to test this (other than failing the travis build like it did just now) 😃

ibrahimlawal:
To be able to test *and ensure* compatibility with the existing one at wire level, I would need some of its sample JSON messages.

ar:
We'll try to take the JSONPackager out of hold a-s-a-p, we also need it for one of our projects too. Will close this one for now but will keep a reference to it.

PR #158, opened 2018-03-20
Add CDATA tag to JsonNode objects when printing

jrfinc:
Also imports Jackson Json library.


jrfinc:
Actually, we don't need the library. Will post a new PR with a better fix.

PR #165, opened 2018-05-03
Target type inference

demsey:
Using type inference in generic methods allows to avoid explicit type casting in code, eg.
```java
      @SuppressWarnings("unchecked")
      Integer someMilis = (Integer) ctx.get(CTX_SOME_TIME);
```
Now it can be written as:
```java
      Integer someMilis = ctx.get(CTX_SOME_TIME);
```

### DIFF
--- a/jpos/src/main/java/org/jpos/security/SecureKeyStore.java
+++ b/jpos/src/main/java/org/jpos/security/SecureKeyStore.java
@@ -62,17 +62,16 @@ public interface SecureKeyStore {
     }
 
 
-
     /**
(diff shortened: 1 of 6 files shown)


ar:
Awesome - thanks for this nice PR!

ar:
FYI, there was a little (but nasty) bug in this PR - fixed in `a9aea1403`

demsey:
Indeed, this is an obvious mistake:(

ar:
But your PR is great, as always, it will save us hundreds of castings.

PR #181, opened 2018-08-01
Changes for support maxage property were added

gregorioosorio:
Please, check if the changes accomplish with the new feature.

As you can see, I proposed the new property: _**maxage**_ instead of **_maxdays_** because is more flexible because this new property is set to seconds.
For example, if you want to set the max age to one week you should set this value:
- maxage = 7*24*3600 = 604800

Also, if you want to disable this new feature just set to zero or a negative number.

### DIFF
--- a/jpos/src/main/java/org/jpos/util/DailyLogListener.java
+++ b/jpos/src/main/java/org/jpos/util/DailyLogListener.java
@@ -98,6 +98,7 @@ public class DailyLogListener extends RotateLogListener{
 		if (maxAge > 0) {
 			maxAge *= 1000;
 		}
+		deleteRegex = cfg.get("delete-regex", defaultDeleteRegex());
 
         setDateFmt(fmt);
         setLastDate(fmt.format(new Date()));
@@ -143,7 +144,12 @@ public class DailyLogListener extends RotateLogListener{
                 rotate = new DailyRotate(), cal.getTime(), sleepTime);
     }
 
-    public synchronized  void logRotate() throws IOException {
+	private String defaultDeleteRegex() {
+		Path prefixPath = Paths.get(prefix);
+		return "^" + prefixPath.getFileName().toString() + ".+\\" + suffix + "\\" + compressedSuffix + "$";
+	}
+
+	public synchronized  void logRotate() throws IOException {
         closeLogFile ();
         super.close ();
         setPrintStream (null);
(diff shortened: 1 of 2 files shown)


gregorioosorio:
@alcarraz thanks for your comments! I will review them. Just notice that the **_MAX_DEPTH_** constant used in **_Files.find()_** function is set to one, that means only the files in that path would be deleted, but you are right, always there is the probability to delete some non-log file.

ar:
I think it's fair enough to check for the file prefix (i.e. `q2-`) and suffix (`.log`).

alcarraz:
@apr I totally agree, but I would implement that as a default regular expression for matching the prefix and suffix , and that can be overwritten by config.

Let's say delete-regex, that way the implementation is pretty much the same but gives more flexibility to the user.

That said I think prefix + suffix is enough and the other can be addressed later if necessary. The DailyLogListener code is too much cumbersome as it is (totally my younger self fault :) ).  

Do you agree @gregorioosorio ?

jpos/src/main/java/org/jpos/util/DailyLogListener.java
```diff
@@ -148,6 +158,38 @@ public synchronized  void logRotate() throws IOException {
...
+
+		if (maxAge <= 0) {
+			logDebug("maxage feature is disabled.");
+			return;
+		}
+
+		Path prefixPath = Paths.get(prefix);
+		Path logNamePath = prefixPath.subpath(prefixPath.getNameCount() - 1, prefixPath.getNameCount());
+
+		String rotatedLogRegex = ".*" + logNamePath.toString() + ".+\\" + suffix + "\\" + compressedSuffix;
```
alcarraz: @gregorioosorio I don't understand the rationale of the ".*", why don't you just match the basename part  of the files in the log directory? this could delete filles that don't start with prefix name. 
I would keep the basename part, and then in the lambda expression check the basename part against the regex instead of the full path, I don't know if I explain myself here

gregorioosorio: @alcarraz with the **".*"** I tried to support the operating system path separator. I agree with you, is simpler to check only with basename part of the file, instead of the whole path, in fact, I will include the characters ^ and $ in the default regex to make sure that match only a file with the exact name. Finally, I will take in count your comment about to include the parameter: _maxage-delete-regex_, which will have the default regex.
I will push the changes shortly! If there is something I am missing, let me know!

alcarraz: I think that would be perfect @gregorioosorio, thank you very much for taking the time to improve this. I was going to take that issue some day, but I don't think I could have done it before next year.

PR #186, opened 2018-09-20
Minus length of length from the 2 byte msg length

chhil:
Currently the channel will wait for 2 additional bytes if the length of length is set to true.
When a length of length is included in the length the getMessageLength should subtract the length of length value to get the bytes to follow. 
This in reference to the changelog entry "New NACChnanel `include-header-length` property [3b5f9d60]"(https://github.com/jpos/jPOS/commit/3b5f9d60d)


ar:
Oh, good catch @chhil !

PR #187, opened 2018-09-24
Update QServer.java

it240884sii:
add ```<send-request>RR</send-request>``` option to QServer. It is allow to send messages to all connected clients in round-robin fashion.


ar:
Nice to have. Thanks for the PR.

PR #188, opened 2018-09-24
It240884sii patch 2

it240884sii:


ar:
Hey, thank you for going the extra mile and updating the doc.

PR #195, opened 2018-12-14
Configurable prepend machine name to log file name

chhil:
Added the ability to prepend the machine name to the log file name to easily identify logs when brought to a central location from multiple servers.
This change is backwards compatible and will prepend the machine based on the configuration in the log listener. By default, prepending machine name is set to false.
An optional configurable separator after the prepended machine name gives some flexibility to naming conventions.


ar:
I understand this has also been implemented in @alwyn's [PR194](https://github.com/jpos/jPOS/pull/195). Should we close this one?

chhil:
Yes please.

-chhil

PR #203, opened 2019-01-31
Add format configuration to field packagers

gitmik:
This change is to enable field packager to use different header (tag and length) formats,
currently some of field packagers use LLLTT or just TTLL.
The formatter allows to pass this information to the packager.


gitmik:
I think I will refactor this code, now I want to reorganize how the new formatter works, it should have references to prefixers and use them to provide the header bytes or unpack into a component, so this will be hidden from the field packager class and it will not duplicate information which is already in prefixers.

ar:
It would help to understand this contribution if you could provide some use cases examples, test cases, or a section in the documentation (whatever is easier for you).

gitmik:
The idea behind this pull request is because the Field Packagers are somewhat configurable with Prefixers, Padders and Interpeters but still the format of the header (the tag and length parts) is fixed.

In my situation I was looking for a LLLTTvvvv packager and couldnt find one between all those fixed ones with name prefix IFE_xxxxx.
I have found ISOTagStringFieldPackager but still the format or order of header parts is fixed and different from my expectations.

This way I implemented a similar packager which takes IsoFieldHeaderFormatter to deal with the ordering of header parts.

The only caveat of the implementation is the case when the header starts with the LENGTH part being followed with TAG so any of LLLTTvvvvvvv, LLTTTvvvvvv or similar.
In this case the LEN calculates the length of tag and the value.
Is this a proper assumption?
The reason is the spec of GICC protocol which puts such a rule.

demsey:
@gitmik Did you check in (Europay packager)(https://github.com/jpos/jPOS/blob/master/jpos/src/dist/cfg/packager/europay.xml)  at _"EUROPAY FIELD 48"_ description?. Where is used `org.jpos.iso.packager.EuroSubFieldPackager` with family of `org.jpos.iso.IFEP_LL<types>`, `org.jpos.iso.IFEP_LLL<types>` - they use `LLTTvvvv` and `LLLTTvvvv` and so on.

I use that case in mastercard banknet protocol where `org.jpos.iso.packager.EuroSubFieldPackager` contains elements from family `org.jpos.iso.IFMC_LL<types>`, `org.jpos.iso.IFMC_LLL<types>` - they use `TTLLvvvv` and `TTLLLvvvv` and so on.

gitmik:
@demsey thanks for suggestion, we already use those packagers in other places, I just couldn't find the exact one for this particular requirement.
I could just add another similar to the IFEMC_LLLCHAR but with following
EbcdicPrefixer.LLL as length prefixer
EbcdicPrefixer.LL as tag 
LeftPadder.ZERO_PADDER
EbcdicInterpreter.INSTANCE 

But just wanted to add something more generic enabling others to specify their own configurations of packagers without adding a new one to the codebase.

demsey:
You're right. I also think that a more general mechanism would be useful. I'm slowly working on it;)

A lot of time ago I took the first step in this direction but it is far from ideal https://github.com/demsey/jPOS/commit/ee3f272235fb9cd9b9dc29e23c48d832adfff852
There are to classes `TagMap` and `Tag` _(they are similar to `TLVList` and `TLVMsg` but works on strings)_. I will update them soon so that they support **any tag size and length size** with the **option to switch order** of `T` with `L` _(they will be use a builer pattern)_.  Similar to the one prepared at https://github.com/demsey/jPOS/commits/tlv-extension2. This will be a good base to do TLV subpackagger.

On the other hand, `EBCDIC` is just character encoding _(in java `'IBM1047'`)_ Maybe inherited _(via `ThreadLocal`)_ attributes on `ISOFieldPackager`s

PR #204, opened 2019-02-11
Add LLLTTvvvvv Ebcdic encoded header and value, length left zero padded.

gitmik:
Due to some discussions over more generic solution in #203 this change will add another option of a LLLTTvvvvv field packager that I finally wanted to create.


gitmik:
@ar sorry I haven't noticed that you merged the #203 
this one is exactly the packager I wanted to use via that change but I didn't want to implement just another option of packager but finally with this PR I thought it may be a faster way to get this working

I dont have a strong opinion whether we should have just #203 or this as well, I will rely on your opinions, still this can be more intuitive for someone working with packagers already.

PR #205, opened 2019-02-14
Add TLVList builder pattern

demsey:
Last commit implements builder pattern for `TLVList`. So is possible parse/unpack TLV byte sequences that are not compliant with ISO/IEC 7816-4 BER-TLV rules. e.g: Visa Base1 Datasets _(which are fixed 2 bytes long)_
```java
    TLVList tl = TLVList.createInstance()
                        .fixedLengthSize(2)
                        .build()
```
After that using `tl` is as usual: unpack, append, pack, etc.
It is also possible to create `TLVList` with a fixed tag size:
```java
    TLVList tl = TLVList.createInstance()
                        .fixedTagSize(2)
                        .build()
```
and even mixing them together.

e.g: when fixed length size 1 is used then length is encoded i 1 byte _(0..255)_ unlike at ISO/IEC 7816-4 only _(0..127)_

The changes affected only `TLVList` and `TLVMsg` _(and corresponding jUnit test)_. They are never used in jPOS itself.
 
#### API changes:
* remove `TLVMsg()`
* remove `TLVMsg.setTag(int)`
* remove `TLVMsg.setValue(byte[] value)`

At problems use just TLVList.append(tag, bytes)

* change `TLVList.unpack` to throws only one unchecked `IllegalArgumentException` instead `ISOException` and extra undocumented  `BufferUnderflowException`
* deprecate `TLVMsg(int, byte)` - use TLVList.append(int, byte) instead.

Most of these changes were necessary to ensure consistency in the situation of the availability of various types of `TLVList` _(which was previously not necessary)_

In addition 2427a9f introduces verification of created tag identifiers.
So now it is not possible to create an illegal _(in ISO/IEC 7816-4 BER-TLV mode)_ `FF`, `00`, '3F', etc. tags. Using them resulted in the creation of an unparsable byte sequence.

### DIFF
--- a/jpos/src/main/java/org/jpos/tlv/TLVList.java
+++ b/jpos/src/main/java/org/jpos/tlv/TLVList.java
@@ -69,9 +69,9 @@ public class TLVList implements Serializable, Loggeable {
      *
      * @param buf raw message
      * @throws ISOException
-     * @throws BufferUnderflowException
+     * @throws IllegalArgumentException
      */
-    public void unpack(byte[] buf) throws ISOException, BufferUnderflowException {
+    public void unpack(byte[] buf) throws ISOException, IllegalArgumentException {
         unpack(buf, 0);
     }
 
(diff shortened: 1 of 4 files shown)


ar:
Awesome work @demsey as usual!

PR #208, opened 2019-03-07
Update com.sleepycat:je to 18.3.12

demsey:
This artifact is available again in maven central (from version 18.1.11 of Jun 13, 2018)
so refference to repository http://download.oracle.com/maven was removed.

ps.
I wonder what is dependent in `jpos/jPOS` on artifacts from http://jpos.org/maven. I can not find anything like that. After removing that reference, the compilation and tests pass.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,7 +29,6 @@ subprojects {
     repositories {
         mavenCentral()
         maven { url 'http://jpos.org/maven' }
-        maven { url 'http://download.oracle.com/maven' }
         mavenLocal()
     }
     if (project.hasProperty("lint")) {
--- a/jpos/libraries.gradle
+++ b/jpos/libraries.gradle
@@ -2,7 +2,7 @@ ext {
     libraries = [
         jdom: 'org.jdom:jdom2:2.0.6',
         jdbm: 'jdbm:jdbm:1.0',
-        sleepycat_je: 'com.sleepycat:je:7.0.6',
+        sleepycat_je: 'com.sleepycat:je:18.3.12',
         commons_cli: 'commons-cli:commons-cli:1.4',
         jline: 'org.jline:jline:3.9.0',
         beanshell: 'org.apache-extras.beanshell:bsh:2.0b6',


ar:
No need to depend on jPOS's Maven in jPOS. It's required in the jPOS-template in order to get nightly snapshot builds, but not in jPOS. 

As for JE, I'm reviewing the release notes for this version, do you have a link handy?

Oracle seems to go up to 7.4.5 (see [release notes](https://www.oracle.com/technetwork/database/berkeleydb/downloads/index-098622.html). Problem with JE updates is that sometimes require a database migration on existing systems, so we may have to hold a bit this PR in order to plan for such upgrades (quite sensitive in production systems).

ar:
Simple smoke test worked out of the box (unbelievable, JE used to require painful migrations). Good!

demsey:
From changelog:
> In JE 7.5 the on-disk file format moved to 15. The file format change is forward compatible in that JE files created with earlier releases can be read when opened with JE 7.5 or later. The change is not backward compatible in that files created with JE 7.5 or later cannot be read by earlier releases. After an existing environment is opened read/write using JE 7.5, the environment can no longer be read by earlier releases.

I understand that files created in version 7.4 and earlier can be read by the new version. They will be converted to format 15, which is not backwards compatible and cannot be read by earlier releases.
The migration itself is automatic but copies of JE files may be useful.

I understand that preparations are necessary. In my applications, JE is used only for queues. In the update conditions there is a record that **queues must be empty**. _(Then possible is even delete JE files)_

ar:
Awesome. We are good to go then. Good to see JE keeps active, it's a great project.

demsey:
WARNING: je in version 18.3.12 by default requires at least 5GB free disk space.
[EnvironmentConfig.html#FREE_DISK](https://docs.oracle.com/cd/E17277_02/html/java/com/sleepycat/je/EnvironmentConfig.html#FREE_DISK)
It can be workaround by place `je.properties` file in space directory _(with `je.freeDisk` property)_

vsalaman:
Yes, this bit me in production the other day. I really think we should start looking at alternatives to 'je'

ar:
Sorry to hear that. Thank you @demsey for raising this issue to our attention. 
@vsalaman Sorry to hear you had the same problem. We already have an alternative, although older, the trusted jdbm Space works pretty well, and of course, we need to look for alternatives, JE is becoming overkill. The good thing about the Spaces is that we can keep older implementations around `jdbm`, `je` and add a new one.

jameshilliard:
I think `je` is also probably causing memory errors in travis(they don't seem to happen every time but are quite frequent).

ar:
They are quite frequent indeed. Why do you think those are related to JE?

jameshilliard:
The memory errors seem to originate from functions within `je`.

ar:
Good point. Will pay attention to those.

PR #215, opened 2019-05-13
SMAdapter keyblock handling extension

demsey:
A set of commits that enables key block support via the `SMAdapter`.

### After applying the commit 36b3e32, the custom HSM driver will not compile.
In order to solve this problem it is required to modify such driver.

In the minimal version, add `<SecureDESKey>` _(to stay with variants LMK handling only)_
``` java
public class SomeCustomHSMDriver extends BaseSMAdapter<SecureDESKey>
```
See file change `JCESecurityModule.java` in 36b3e32

In order to fully operate the key block by `SomeCustomHSMDriver` is required to add
``` java
public class SomeCustomHSMDriver extends BaseSMAdapter<SecureKey>
```
adjust _(to support key block)_ all used by `SomeCustomHSMDriver` methods and if required implement some new version of methods:
  * generateKey
  * importKey
  * exportKey
  * translateKeyFromOldLMK
  * generateKeyPair

To use `ExtKeyUsage` from b46ccaa with proprietary specific key usages is required to define custom resource `META-INF/org/jpos/security/proprietary-hsm.properties`. For example see at
`jpos/src/test/resources/META-INF/org/jpos/security/proprietary-hsm.properties`

### DIFF
--- a/jpos/src/main/java/org/jpos/security/ModeOfUse.java
+++ b/jpos/src/main/java/org/jpos/security/ModeOfUse.java
@@ -1,0 +1,143 @@
+/*
+ * jPOS Project [http://jpos.org]
+ * Copyright (C) 2000-2019 jPOS Software SRL
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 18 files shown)


jpos/src/main/java/org/jpos/security/ModeOfUse.java
```diff
@@ -0,0 +1,143 @@
...
+    /**
+     * The key may be used to perform both encrypt and decrypt operations.
+     */
+    ENCDEC          ('B', "Encryption and Decryption"),
+
+    /**
+     * The key may be used to perform MAC calculation <i>(both generate &
+     * verify)</i> operations.
+     */
+    GENVER          ('C', "Verification and Genaration of MAC, CVD"),
```
ar: Watch out super minor typo (Generation)

demsey: Heh I have to write "Feel free to submit your comments"

PR #217, opened 2019-06-01
Make AsciiPrefixer fail if input doesn't contain digits

jmones:
`AsciiPrefixer.decodeLength()` accepted input characters that were not digits. It also produced a bad length as a result. This is a dangerous situation, as the behaviour of the parser can then not be predicted. This pull request tried to keep style and algorithm, but justs throws an exception if input characters are not correct.

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/AsciiPrefixer.java
+++ b/jpos/src/main/java/org/jpos/iso/AsciiPrefixer.java
@@ -86,12 +86,16 @@ public class AsciiPrefixer implements Prefixer
     }
 
     @Override
-    public int decodeLength(byte[] b, int offset)
-    {
+    public int decodeLength(byte[] b, int offset) throws ISOException {
         int len = 0;
         for (int i = 0; i < nDigits; i++)
         {
-            len = len * 10 + b[offset + i] - (byte)'0';
+            byte d = b[offset + i];
+            if(d < '0' || d > '9')
+            {
+                throw new ISOException("Invalid character found. Expected digit.");
+            }
+            len = len * 10 + d - (byte)'0';
         }
         return len;
     }
--- a/jpos/src/test/java/org/jpos/iso/AsciiPrefixer2Test.java
+++ b/jpos/src/test/java/org/jpos/iso/AsciiPrefixer2Test.java
@@ -35,8 +35,12 @@ public class AsciiPrefixer2Test {
     @Test
     public void testDecodeLength() throws Throwable {
         byte[] bytes = new byte[2];
-        int result = AsciiPrefixer.LL.decodeLength(bytes, 0);
-        assertEquals("result", -528, result);
+        try {
+            int result = AsciiPrefixer.LL.decodeLength(bytes, 0);
+            fail("Expected ISOException to be thrown");
+        } catch (ISOException ex) {
+            assertEquals("ex.getMessage()", "Invalid character found. Expected digit.", ex.getMessage());
+        }
     }
 
     @Test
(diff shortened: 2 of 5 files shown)


jmones:
An additional note: later on I realized that other prefixers (such as `BcdPrefixer` or `EbcdicPrefixer`) have similar issue. So I wonder if this is designed intentionally like this to avoid a pair of CPU cycles. I think that parsers should stop as soon as possible, because it's difficult to predict if some clever crafted input may end up with a sucessfully parsed input.

jpos/src/main/java/org/jpos/iso/AsciiPrefixer.java
```diff
@@ -86,12 +86,16 @@ public void encodeLength(int length, byte[] b) throws ISOException
...
     @Override
-    public int decodeLength(byte[] b, int offset)
-    {
+    public int decodeLength(byte[] b, int offset) throws ISOException {
         int len = 0;
         for (int i = 0; i < nDigits; i++)
         {
-            len = len * 10 + b[offset + i] - (byte)'0';
+            byte d = b[offset + i];
+            if(d < '0' || d > '9')
```
ar: Should we use `Character.isDigit(d)` instead to better show the intention? 
(although it's pretty clear in your code).

jmones: Hi Alejandro, thx for reactivity.

I think it's slightly better as it is.

Please note that Character.isDigit() allows not only ASCII digits but also:
- '\u0030' through '\u0039', ISO-LATIN-1 digits ('0' through '9')
- '\u0660' through '\u0669', Arabic-Indic digits
- '\u06F0' through '\u06F9', Extended Arabic-Indic digits
- '\u0966' through '\u096F', Devanagari digits
- '\uFF10' through '\uFF19', Fullwidth digits

That's probably not an issue while using it together with the rest of the system (because byte to string is hardcoded to use ISO-8859-1 encoding), but I think it's a good idea to be restrictive in the case someone uses the class in isolation.

BTW: also comment that in some JAVADOC it is stated that US_ASCII (assumed 7 bit) is used, but the library uses extended 8-bit character set in ISO-8859-1 codepage.

alcarraz: Well I don't have any inclination towards one way or the other, but since d is a byte not a char it can never be grater than '\u00FF'.

ar: Both good points. No need to change it, will merge it as it is.

jmones: @alcarraz so true! My comment was pointless then... I really don't care which option.
Finally: what about BcdPrefixer or EbcdicPrefixer? Do you plan to patch it? Do you want inputs on those?

ar: Beware of side-effects with BCD related stuff and some special characters used in track2 ('D')

alcarraz: I think @jmones is talking only about the prefixers, so it shouldn't affect the interpreters

ar: Absolutely, I understand. I was just raising a flag and self reminder in case the fix could be attempted at the `ISOUtil` level.

jmones:
Hi, again. A colleague pointed me a out a much better/simplier solution:

```
public int decodeLength(byte[] b, int offset)
{
  try {
    return Integer.parseInt(new String(b, offset, nDigits, Charsets.ISO_8859_1));
  } catch (NumberFormatException e) {
    throw new ISOException("failed to decoded length", e);
  }
}
```

alcarraz:
looks cleaner and simpler but is more expensive since it involves the creation of a string object and that is a heavily used operation. One of the many good things of jpos is how fast it is for parsing/packaging.

PR #221, opened 2019-06-29
Run tests against jdk 11 and 12 while allowing failures.

jameshilliard:
This approach avoids disabling any tests like #220 does while still running the build against jdk 11 and 12(but with failures for those ignored).

This should make it easier to incrementally fix the tests under jdk 11 and jdk 12.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,19 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: java
 jdk:
-- oraclejdk8
+  - oraclejdk8
+  - openjdk8
+  - openjdk11
+  - openjdk12
 before_script:
-- echo $JAVA_OPTS
-- export JAVA_OPTS=-Xmx4G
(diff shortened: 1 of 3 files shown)


ar:
FYI, just forced the tests to run again. Passing alright this time.
https://travis-ci.org/jpos/jPOS/builds/552500017

ar:
I see you removed tests against oraclejdk8, while people are fading away from it, the truth is most production sites are using it. I'm afraid we need to keep it around for a while.

jameshilliard:
I removed oraclejdk since it it's no longer available for automated download.

ar:
Well.. it's still working in `master` build

jameshilliard:
I've noticed that...guessing it's cached.

jameshilliard:
Force pushed restoring it.

jameshilliard:
going to close this in favor of #223

PR #222, opened 2019-07-01
Optimize Context.dumpMap to avoid locking during dumpEntry

jameshilliard:
This should fix an issue where testConcurrentException is extremely slow on some systems.

### DIFF
--- a/jpos/src/main/java/org/jpos/transaction/Context.java
+++ b/jpos/src/main/java/org/jpos/transaction/Context.java
@@ -247,9 +247,12 @@ public class Context implements Externalizable, Loggeable, Pausable, Cloneable {
 
     protected void dumpMap (PrintStream p, String indent) {
         if (map != null) {
-             synchronized(map) {
-                map.entrySet().forEach(e -> dumpEntry(p, indent, e));
-             }
+            Map<Object,Object> cloned;
+            cloned = Collections.synchronizedMap (new LinkedHashMap<>());
+            synchronized(map) {
+                cloned.putAll(map);
+            }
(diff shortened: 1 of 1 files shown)


ar:
While I like your PR at first sight, I need to think about the side effects of this, specially in situations where the Context is forwarded to a secondary TransactionManager. While the map is cloned in our PR, the Map entries are not. We had a hard time nailing down some concurrent map modifications regarding this method, and I believe your PR is good and won't have side effects, but I'm still reluctant to blindly merging it until I give it more thought. 

Thank you for the PR.

ar:
May I ask the rationale for this PR? 
Have you experienced a performance problem with the dump? 
(just want to get some context).

jameshilliard:
> Have you experienced a performance problem with the dump?

Yes, the performance issue was bad enough that test appears to be frozen without this on one of my systems.

ar:
Got it. Do you remember which test in particular? Context.dump (when dumped through the TransactionManager) shouldn't cause any contention. Some artificial tests may, though.

jameshilliard:
It is testConcurrentException.

ar:
Got it. I wrote that one trying to reproduce that ConcurrentException that wasn't easy to reproduce, that's why I had to be a bit aggressive with that test. Glad your fix doesn't break that test, that's good.

jameshilliard:
> Glad your fix doesn't break that test, that's good.

Yeah, from my testing that test spent most of the time with `dumpEntry` blocking `ctx.put("Prop-" + i, i)` in the `addEvent` thread without this fix.

PR #223, opened 2019-07-01
Run tests against all recent jdk versions.

jameshilliard:
This includes #222 .

Most test failures were just due to exception message syntax changes which is straightforward to handle without breaking compatibility with jdk8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,35 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: java
 jdk:
-- oraclejdk8
+  - openjdk8
+  - openjdk9
+  - openjdk10
+  - openjdk11
+  - openjdk12
 before_script:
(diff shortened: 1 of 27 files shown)


ar:
I see this PR includes PR #222 that is still under review. No problem, but it's usually a good idea to keep them isolated (easier to maintain).

jameshilliard:
rebased on master

jameshilliard:
I have all tests passing [here](https://travis-ci.com/jameshilliard/jPOS/builds/117745210).

ar:
Just forced a restart, but I'm afraid it's still failing.

jameshilliard:
So `oraclejdk8` installation appears to be broken on `xenial` but not `trusty`, so I've overridden it to run on `trusty` while other tests all run under `xenial`. This seems to be a known travis issue.

PR #224, opened 2019-07-02
new property allow-extra-fields for CheckFields

alcarraz:
This let's use the CheckFields participant for controlling mandatory fields, and parse some optional ones, but no need to define all possible optionals.

Also changed a StringBuffer for a StringBuilder for a local variable


ar:
Although I like to configure which fields are optional (in order to early-detect possible version mismatches among endpoints) I can see how this property can come handy. Thank you!

PR #226, opened 2019-07-02
Lower default je cache percentage and make it configurable.

jameshilliard:
This is an attempt at fixing the frequent `java.lang.OutOfMemoryError` errors in travis.


jameshilliard:
Closing in favor of #227.

PR #227, opened 2019-07-02
Disable LOG_DETECT_FILE_DELETE

jameshilliard:
This appears to be the source of `java.lang.OutOfMemoryError` errors in travis.

It doesn't seem to be worth having this enabled due to excessive resource usage.

### DIFF
--- a/jpos/src/main/java/org/jpos/space/JESpace.java
+++ b/jpos/src/main/java/org/jpos/space/JESpace.java
@@ -26,6 +26,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.Semaphore;
 
 import com.sleepycat.je.*;
+import com.sleepycat.je.config.EnvironmentParams;
 import com.sleepycat.persist.EntityStore; 
 import com.sleepycat.persist.StoreConfig; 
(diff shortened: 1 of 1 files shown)


jameshilliard:
So not really sure this is the correct fix as I still seem to get some `java.lang.OutOfMemoryError` errors, although they seem somewhat less frequent.

ar:
How do you link this property to the OutOfMemory errors? I believe log file deletion is a good to have property, perhaps we could change the LOG_DETECT_FILE_DELETE_INTERVAL, but I'd like to know why you think this is related to the OutOfMemory errors (shouldn't be too memory consuming, unless there's a BerkeleyDB bug).

jameshilliard:
I guessed at that because a lot of the travis memory errors were hitting threads within the codepaths this flag enables/disables. Seems this probably isn't the root cause however.

ar:
I'm not keen to change configurations like this, that have been working alright for years in jPOS production sites, just to blind guess if that is causing a Travis error. I prefer a Travis error rather than a production error.

jameshilliard:
Yeah, at first I was getting tests all passing with this option but then started seeing errors again after rerunning so probably something else then.

PR #231, opened 2019-07-12
Attempt to fix java.lang.OutOfMemoryError.

jameshilliard:
Not sure if I caught all the sources of these errors but these changes seem to reduce how often they occur. I've been completely unable to reproduce any of these `java.lang.OutOfMemoryError` failures locally and am assuming that the main environment difference is that travis has very little CPU power allocated to run background garbage collection unlike my development systems.

It also seems to be nearly impossible to actually terminate all JESpace background threads and clean up resources properly, as a workaround I have those tests now running in isolated processes(#232 is that change separated out for reference).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,6 @@ jdk:
   - openjdk10
   - openjdk11
   - openjdk12
-before_script:
-  - echo $JAVA_OPTS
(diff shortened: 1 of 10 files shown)


jameshilliard:
Closing as this is likely fixed in #334 or earlier.

PR #232, opened 2019-07-12
Ensure tests using JESpace are run as separate processes.

jameshilliard:
This might be enough by itself to fix most of the `java.lang.OutOfMemoryError` errors. I was attempting to find a way to kill the background threads in #231 but couldn't find a way to do that(that PR also includes this change).


jameshilliard:
Closing as this is likely fixed in #334.

PR #235, opened 2019-07-23
Add space push and out expiration tests

jameshilliard:


ar:
FYI 
```
Users/apr/git/jpos/jpos/src/test/java/org/jpos/space/JESpaceTest.java:24: error: cannot find symbol
import org.junit.Test;
                ^
  symbol:   class Test
  location: package org.junit
/Users/apr/git/jpos/jpos/src/test/java/org/jpos/space/JESpaceTest.java:29: error: cannot find symbol
    @Test
     ^
 
```

jpos/src/test/java/org/jpos/space/JDBMSpaceTestCase.java
```diff
@@ -155,6 +155,36 @@ public void testPush() {
         assertNull (sp.rdp ("PUSH"));
     }
     @Test
+    public void testOutExpire() {
+        sp.out ("PUSH", "ONE", 1000L);
```
ar: Not a big deal, key name can be anything, but "PUSH" in the `testOutExpire` is probably confusing to a new user. Perhaps you want to use "OUT" as the key, or any other key.

jameshilliard:
> FYI
> 
> ```
> Users/apr/git/jpos/jpos/src/test/java/org/jpos/space/JESpaceTest.java:24: error: cannot find symbol
> import org.junit.Test;
>                 ^
>   symbol:   class Test
>   location: package org.junit
> /Users/apr/git/jpos/jpos/src/test/java/org/jpos/space/JESpaceTest.java:29: error: cannot find symbol
>     @Test
>      ^
>  
> ```

There shouldn't be any `import org.junit.Test;` imports at all since those were migrated to junit5.

jpos/src/test/java/org/jpos/space/JDBMSpaceTestCase.java
jameshilliard: right, will change

ar:
> There shouldn't be any import org.junit.Test; imports at all since those were migrated to junit5.
You're right. It was a local file in my tree that wasn't migrated.

Issue #237, opened 2019-08-07
SimpleConfigurationTest.testReadEnvironmentVariable test fails on MS Windows Environments.


dbergert:
SimpleConfigurationTest.testReadEnvironmentVariable test fails on MS Windows Environments, due to MS Windows Systems not having a "HOME" env varible which is used in this test.

ar:
Thank you for the bug report and PR. Just merged it.

PR #238, opened 2019-08-07
Address #237 SimpleConfigurationTest.testReadEnvironmentVariable Fails on MS Windows

dbergert:
Address #237 SimpleConfigurationTest.testReadEnvironmentVariable Fails on MS Windows


ar:
Thank you Dave!

PR #257, opened 2019-10-23
Update email address in the credits file.

fgonzal:


ar:
FYI, applied manually in 2f103b269 to avoid changes in other users (encoding stuff, look at your diff)

PR #178, opened 2018-07-19
Kafka log listener

andresantoniuk:
This is a preliminary version of a new LogListener  this goals:
- Use Kafka as a way to  "concentrate" logs from different jPOS modules/instances/servers
- Events formated as jSON so it's is easy to ingest this events to Elastic via [Logstash Kafka Input](https://www.elastic.co/guide/en/logstash/current/plugins-inputs-kafka.html)

Configuration is done via a new _log-listener_ section on `00_logger.xml` that forwards configuration options to the [Kafka Producer API](https://kafka.apache.org/11/javadoc/index.html?org/apache/kafka/clients/producer/KafkaProducer.html)

```
  <log-listener class="org.jpos.util.KafkaLogListener">
    <property name="topic" value="jpos" />
    <property name="KafkaProducer.bootstrap.servers" value="localhost:9092" />
    <property name="KafkaProducer.acks" value="all" />
    <property name="KafkaProducer.key.serializer" value="org.apache.kafka.common.serialization.StringSerializer" />
    <property name="KafkaProducer.value.serializer" value="org.apache.kafka.common.serialization.StringSerializer" />
    <property name="KafkaProducer.retries" value="0" />
    <property name="KafkaProducer.batch.size" value="16384" />
    <property name="KafkaProducer.linger.ms" value="1" />
    <property name="KafkaProducer.buffer.memory" value="33554432" />
  </log-listener>
```

### Testing environment
- Kafka setup following https://kafka.apache.org/quickstart
- Logstash pipeline

```
input {  
  kafka {
    bootstrap_servers => "localhost:9092"
    topics => ["jpos"]
  }
}
filter {
  json {
    source => "message"
  }
  mutate {
    remove_field => [ "message" ]
  }
}
output {
  stdout { 
    codec => rubydebug
  }
  elasticsearch{
    hosts => ["localhost:9200"]
    index => "jposlog-%{+YYYY.MM.dd.hh}"
  }
}
```

- Sample log data in elastic

```
GET jposlog-2018*/_search

{
  "took": 1134,
  "timed_out": false,
  "_shards": {
    "total": 10,
    "successful": 10,
    "skipped": 0,
    "failed": 0
  },
  "hits": {
    "total": 206,
    "max_score": 1,
    "hits": [
      {
        "_index": "jposlog-2018.07.19.01",
        "_type": "doc",
        "_id": "nT8QsGQBU4jSE_wmPlfZ",
        "_score": 1,
        "_source": {
          "IP": "52.7.83.125",
[...]

PR #258, opened 2019-10-25
Enable JSON Logging

erlanggaelfallujah:
Modified jpos log to support json format logging. Case if we shipping the logs file to a log aggregator and using json for standar message passing.

Below add format configuration (default format is xml) :

```
<log-listener class="org.jpos.util.DailyLogListener">
        <property name="format" value="json" />
        <property name="window" value="86400" />
        <property name="prefix" value="log/q2-json" />
        <property name="suffix" value=".log"/>
</log-listener>
```

Example Logs :

```
{
  "log": {
    "realm": "",
    "at": "2019-10-25T09:57:01.382233",
    "unpack": {
      "bitmap": "{3, 11, 24, 41, 46, 47, 48, 61}",
      "unpack": [
        {
          "packager": "org.jpos.iso.IFB_NUMERIC",
          "value": 450000,
          "fld": 3
        },
        {
          "packager": "org.jpos.iso.IFB_LLLBINARY",
          "value": {
            "type": "binary",
            "content": "DF90030"
          },
          "fld": 46
        }
      ]
    }
  }
}
{
  "log": {
    "realm": "",
    "at": "2019-10-25T09:57:01.377541",
    "error": "com.ranggalabs.swc.exception.Exception: UnLog at org.jpos.q2.iso.MUXPool.request(MUXPool.java:79) at com.ranggalabs.swc.Switcher.SessionManagerBean.sessionManaging(SessionManagerBean.java:80) "
  }
}
{
  "log": {
    "realm": "",
    "at": "2019-10-25T09:57:01.351143",
    "unpack": {
      "bitmap": "{3, 11, 24, 41, 46, 47, 48, 61}",
      "unpack": [
        {
          "packager": "org.jpos.iso.IFB_NUMERIC",
          "value": 450000,
          "fld": 3
        },
        {
          "packager": "org.jpos.iso.IFB_LLLBINARY",
          "value": {
            "type": "binary",
            "content": "DF90030"
          },
          "fld": 46
        }
      ]
    }
  }
}
{
  "log": {
    "realm": "",
    "at": "2019-10-25T09:57:01.361113",
    "lifespan": "4 ms",
    "receive": {
      "exception": {
        "sqlexception": "null",
        "sqlstate": "null",
        "vendorerror": "0",
        "stacktrace": [
          "java.sql.SQLException",
          "at org.jpos.util.JsonRotateListenerTest.testAddMessageAndSqlException(JsonRotateListenerTest.java:147)",
          "at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)",
          "at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)",
          "at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)",
          "at java.base/java.lang.reflect.Method.invoke(Method.java:567)",
          "at org.gradle.internal.concurrent.ThreadFactoryImpl$ManagedThreadRunnable.run(ThreadFactoryImpl.java:56)",
          "at java.base/java.lang.Thread.run(Thread.java:835)"
        ]
      }
    }
  }
}
{
  "log": {
    "realm": "",
    "at": "2019-10-25T09:57:01.308615",
    "info": {
      "isomsg": {
        "field": [
          {
            "id": 0,
            "value": "0800"
          },
          {
            "id": 3,
            "value": 990000
          },
          {
            "id": 7,
            "value": 101915
          }
        ],
        "header": "49534F303",
        "direction": "outgoing"
      }
    }
  }
}
```

### DIFF
--- a/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
+++ b/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
@@ -23,8 +23,6 @@ import java.util.stream.Stream;
  */
 public class JSONLogEvent implements BaseLogEvent {
 
-    private static final DateTimeFormatter formatter = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SSS");
-
     @Override
     public String dumpHeader(PrintStream p, String indent, String realm, Instant dumpedAt, Instant createdAt, boolean noArmor) {
         if (noArmor) {
--- a/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
+++ b/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
@@ -93,6 +93,7 @@ public class JSONLogEvent implements BaseLogEvent {
                         } else if (o instanceof SQLException) {
                             p.print("SQLException");
                         } else if (o instanceof Throwable) {
+                            // TODO: Fetch n lines
                             p.print("{ \"exception\" : { \"name\":\"" + ((Throwable) o).getMessage()+"\",");
                             p.print("\"stackTrace\":\"");
                             p.print(Arrays.toString(((Throwable)o).getStackTrace()));
@@ -106,10 +107,6 @@ public class JSONLogEvent implements BaseLogEvent {
                             p.print("Element");
                         } else if (o != null) {
                             if(stringBuilder.length()==0){
-                                /*
-                                    for handling
-                                    "connect":"Try 0 127.0.0.1:6990  Connection refused (Connection refused)Unable to connect"}}
-                                 */
                                 isOpenQuote = true;
                                 stringBuilder.append("\"");
                             }
--- a/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
+++ b/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
@@ -55,7 +55,7 @@ public class JSONLogEvent implements BaseLogEvent {
     @Override
     public void dumpTrailer(PrintStream p, String indent, boolean noArmor) {
         if (!noArmor) {
-            p.print ("\n}\n");
+            p.print (indent+"\n}\n");
         }
     }
 
--- a/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
+++ b/jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
@@ -23,9 +23,10 @@ import java.util.stream.Stream;
  */
 public class JSONLogEvent implements BaseLogEvent {
 
-    public static final String XML_TAG_PATTERN = "(?s).*(<(\\w+)[^>]*>.*</\\2>|<(\\w+)[^>]*/>).*";
+    private static final String XML_TAG_PATTERN = "(?s).*(<(\\w+)[^>]*>.*</\\2>|<(\\w+)[^>]*/>).*";
     private static final String STACK_TRACE_TAG_PATTERN = "(^\\d+\\) .+)|(^.+Exception: .+)|(^\\s+at .+)|(^\\s+... \\d+ more)|(^\\s*Caused by:.+)";
-    public static final Pattern STACK_TRACE_REGEX = Pattern.compile(STACK_TRACE_TAG_PATTERN, Pattern.MULTILINE);
+
+    private static final Pattern STACK_TRACE_REGEX = Pattern.compile(STACK_TRACE_TAG_PATTERN, Pattern.MULTILINE);
 
     @Override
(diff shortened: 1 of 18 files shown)


andresantoniuk:
I tried something similar in https://github.com/jpos/jPOS/pull/178 but I didn't manage to invest more time on that; hope this move forward as it's a really important improvement!

mikesoares:
This is great! A few people from our team here at Zero Financial are currently looking this over.

ar:
Great work. I'll review in detail, but have a few concerns:

- I fear this would break existing custom log listeners and filters (may be not)
- This substantial PR may have IP related issues (no copyright, no CLA nor CCLA)
- I see some internal company information leaked into the code in some tests
- Some System.out.printlns here and there that we try not to have in jPOS
- Some internal heavy used methods (such as log indenting) now require massive object creation and iterations on StringBuilders. I fear that code like this could have some performance issues, we are into the business of improving performance, not making it slower.
- Adds new dependency on old JSON libraries. We have the JSONPackager on hold until as a community decide which new JSON dependency we plan to use.

The jPOS Logger subsystem (I like to call it 'Audit subsystem' because it's not the same thing as traditional loggers such as log4j/j.u.l/slf4j/logback) is looking for some deep changes and improvements. Adding this PR at this time would perpetuate some of the faults we have.

But it is a great PR that raises the priority of deeper changes in the Logger subsystem, that I hope we can add during 2.1.5.

jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
```diff
@@ -0,0 +1,211 @@
...
+                sb.append (" ms\"");
+            }
+
+            p.print(sb.toString());
+        }
+        return indent;
+    }
+
+    @Override
+    public void dumpTrailer(PrintStream p, String indent, boolean noArmor) {
```
richard-li-zero: 'indent' is not used in this method

jpos/src/main/java/org/jpos/util/log/event/XMLLogEvent.java
```diff
@@ -0,0 +1,130 @@
...
+                sb.append ("ms\"");
+            }
+            sb.append ('>');
+            p.println (sb.toString());
+        }
+        return indent + "  ";
+    }
+
+    @Override
+    public void dumpTrailer(PrintStream p, String indent, boolean noArmor) {
```
richard-li-zero: same as above.

jpos/src/main/java/org/jpos/util/log/event/JSONLogEvent.java
erlanggaelfallujah: updated, now is used.

jpos/src/main/java/org/jpos/util/log/event/XMLLogEvent.java
erlanggaelfallujah: It's legacy code, but it's used 

`p.println (indent + "</log>");`

erlanggaelfallujah:
I was improve on indent process

erlanggaelfallujah:
Hi @ar ,

- **I fear this would break existing custom log listeners and filters (may be not)**
Yeah, i think we need another case testing. And maybe, my unit test not covered this case.

- **This substantial PR may have IP related issues (no copyright, no CLA nor CCLA)**
Please guide me for this issues, because this first contribution to open source project.

- **Some internal heavy used methods (such as log indenting) now require massive object creation and iterations on StringBuilders. I fear that code like this could have some performance issues, we are into the business of improving performance, not making it slower.**
I was improved log indenting process, and need review again for other issues.

- **Adds new dependency on old JSON libraries. We have the JSONPackager on hold until as a community decide which new JSON dependency we plan to use.**
I was tried two dependencies (org.json and com.fasterxml.jackson). Jackson dependency have limitation, one of them the case that i found. If tag has duplicate keys they will get overwritten. So for alternative i am choose using org.json.

ar:
I really appreciate the effort you put in this PR, but I believe this same goal can be achieved in a less intrusive way, implementing it as just a LogListener, perhaps added to the jPOS-EE project where different teams can start using the functionality, testing it performance, fixing problems, adding features, and then once battle tested perhaps graduate it to core jPOS. That's the way we usually do with most stuff. I suggest you follow-up the discussion in the jPOS slack, or the jPOS users mailing list. That said, I'll keep this PR open and perhaps we can merge it after 2.1.4 if the other route (log listener) is not feasible.

ar:
Thank you for this contribution. closing it but we'll keep it in mind as a reference.

The jPOS Logger/Audit subsystem helped us well for over 20 years. Things are changing now, we are aware of jPOS systems running in over 500 nodes that generate logs that need to be aggregated. I don't think an adapter from the verbose XML logs into a JSON representation is the way to go. We hope to be able to provide a deeper and new approach to this problem.

PR #271, opened 2020-01-18
PosDataCode: safety length/null checks & safety copy in byte[] constr…

barspi:
…uctor

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/PosDataCode.java
+++ b/jpos/src/main/java/org/jpos/iso/PosDataCode.java
@@ -204,7 +204,7 @@ public class PosDataCode implements Loggeable {
         b[15] = (byte) (securityCharacteristic >>> 24);
     }
 
-    public PosDataCode (byte[] b) {
+    private PosDataCode (byte[] b) {
         if (b != null) {
             // will always use our own internal copy of array
             int copyLen= Math.min(b.length, 16);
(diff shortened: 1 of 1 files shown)


jpos/src/main/java/org/jpos/iso/PosDataCode.java
```diff
@@ -205,8 +204,12 @@ public PosDataCode (
         b[15] = (byte) (securityCharacteristic >>> 24);
     }
 
-    private PosDataCode (byte[] b) {
-        this.b = b;
+    public PosDataCode (byte[] b) {
```
ar: Why are you changing this constructor to public? We have a `PosDataCode.valueOf(byte[])` factory method. I'm OK with the null validation, but don't understand why we need the public constructor.

barspi: Hmm this is something I brought over from the PosDataCode in a custom jCard which, I think, doesn’t have valueOf, and the byte[] constructor is public. I didn’t notice it’s  private in this version.
If we have an alternative public way then it’s ok. My idea was to have only one version of the enum around, and we can deprecate/eliminate the one in jCard.

ar: PosDataCode was initially implemented in jCard, then improved and graduated to jPOS. At some point we'll start using the latter in jCard too. What problem are you trying to solve? NPE check in constructor?

barspi: In particular I was trying to avoid the NPE when constructing (and having to check the null outside in the client code which is noisy)
Also, having in the same project two versions of a class with the same name and purpose (in different packages) gave me a few minutes of head-scratching because of the errors in the IDE until I figured out what was going on.

PR #272, opened 2020-01-22
Added helper methods to Environment to load properties and replace th…

alcarraz:
…eir values


jpos/src/main/java/org/jpos/core/Environment.java
```diff
@@ -235,4 +235,57 @@ public String toString() {
...
+
+    /**
+     * Helper method to replace {@link Properties} values according to
+     * {@link #getProperty(String)} syntax
+     * @param props {@link Properties} to replace.
+     */
+    public static void replaceValues(Properties props) {
+        for (Map.Entry entry : props.entrySet()) {
+            String k = (String) entry.getKey();
+            String v = get((String) entry.getValue(), "");
```
ar: The standard "get(String)" would return the String as the default. See https://github.com/jpos/jPOS/blob/master/jpos/src/main/java/org/jpos/core/Environment.java#L70&L72

I found that useful to detect unconfigured properties (the returned `${...}` was a good eye-catcher).

Should we use `""` as the default here (in your line 287) or just use `get((String) entry.getValue())` ?

ar: And BTW, you could call `getEnvironment()` at the beginning of the loop and then call its `getProperty` method, just to speed up a bit (I know it's probably not required on small config files, but I come from 8bit computing ... :)

PR #273, opened 2020-01-22
Concept of mapping functions along with a RemoveNew…

alwyn:
…LinesMapper implementatios that can be used by a MappingLogEventWriter.  Additionally the addition of a FlatLogListener to achieve similar result without using the MappingLogEventWriter.

### Goal

Provide the ability to modify LogEvents before they are dumped and the output of a dumped LogEvent before it is written to the output stream.

### Design

Introduce the concept of Mappers that extend Java's Function interface.
These mappers map from a version of one type to a modified version of the same type.

E.g. **LogEventMapper** maps a LogEvent to a modified LogEvent.
**ByteArrayMapper** maps a byte[] to a modified byte[]

These mappers can then be chained in a LogEventWriter to allow flexible control over log output.

### Implementation

A **MappingLogEventWriter** is added that can be configured with multiple event-mappers and output-mappers.  The event-mappers are LogEventMappers for pre-dump modification of LogEvents.  The output-mappers are ByteArrayMappers for post-dump modification of the output stream.

In the event that output-mappers are defined, the original PrintStream is hijacked to capture the dumped data, the data is modified and then replayed to the intended target PrintStream.

**NB.**  Note that a log-listener that uses a MappingLogEventWriter to modify LogEvents will pass the changes on to subsequent log-listeners since the LoggerAdaptor passes the processed LogEvents from listener to listener.

The **RemoveNewLinesMapper** has been added as an output mapper to remove newlines and optionally consolidate multiple spaces into a single space.  By default this mapper will add a newline at the end of the LogEvent after all have been removed.  This mapper is important in Cloud environments where logparsers have difficulty in determining where a log message starts and where it ends.

All the mappers have been added to a new package **org.jpos.util.function** since the expectation is that there will be more core functionality that could be implemented as functions.

### FlatLogListener

Additionally **FlatLogListener** has been added to provide the **RemoveNewLinesMapper** functionality without using the **MappingLogEventWriter**.  Note that using this listener will impact all subsequently defined LogListeners in the LoggerAdaptor configuration.


jpos/src/main/java/org/jpos/util/function/RemoveNewLinesMapper.java
```diff
@@ -0,0 +1,92 @@
...
+ * <b>NB. Do not set combine-spaces to true if you have data where spaces are significant.</b>
+ *
+ * @author Alwyn Schoeman
+ * @since 2.1.4
+ */
+public class RemoveNewLinesMapper implements ByteArrayMapper, Configurable {
+    boolean combineSpaces = false;
+    boolean newLineAtEnd = true;
+    final byte SPACE = ' ';
+    final byte NEWLINE = '\n';
```
chhil: How about using a  System.lineSeparator() instead of a hardcoded '\n'

alwyn: Very good idea, I will test it out.

Also I need to make sure this works as expected on Windows systems, which I omitted.

alwyn: Updated with latest commit

PR #288, opened 2020-03-18
Run tests under openjdk 14

jameshilliard:
May want to wait on merging this until there's a final release for gradle 6.3(it's currently on release candidate 3) which is required for Java 14 support but tests all seem to be passing under Java 14 at least.


jameshilliard:
This should be good to go now, gradle 6.3 was released.

jameshilliard:
rebased

Issue #289, opened 2020-03-23
Q2 log shows deploy file renamed to bad but its not actually renamed.


chhil:
https://github.com/jpos/jPOS/blob/master/jpos/src/main/java/org/jpos/q2/Q2.java#L932-L939

The java File.renameTo returns a boolean, and we should use that flag and log the appropriate error if it was renamed or an attempt to rename failed.

ar:
Fixed in https://github.com/jpos/jPOS/pull/290

PR #290, opened 2020-03-23
Check if rename file succeeded  or failed

chhil:
https://github.com/jpos/jPOS/issues/289
Check if rename succeeded and log appropriate message to Q2 log.


ar:
Good catch @chhil - I can see this can easily happen when deploy has permissions issues.

Issue #311, opened 2020-05-26
The Q2 based system stops working through TimerTask


demsey:
It is quite easy to stop the system by using `DefaultTimer.schedule()`. The `TimerTask` passed there cannot throw any exceptions or errors _(e.g. `SpaceError`)_.
It is documented in [Timer javadoc](https://docs.oracle.com/javase/8/docs/api/java/util/Timer.html)

> If the timer's task execution thread terminates unexpectedly, for example, because its stop method is invoked, any further attempt to schedule a task on the timer will result in an IllegalStateException, as if the timer's cancel method had been invoked.

As example the JDMBSpace implement TimerTasks
```java
    public void run () {
        try {
            gc();
        } catch (Exception e) {
            e.printStackTrace(); // this should never happen
        }
    }
```
but at the end of the `gc()` method we have ... `SpaceError`
```java

    public void gc () {
        final String GCKEY = "GC$" + Integer.toString (hashCode());
        final long TIMEOUT = 24 * 3600 * 1000;
        Object obj;
        try {
            synchronized (this) {
                ...
            }

            ...
        } catch (IOException e) {
            throw new SpaceError (e);
        }
    }
```
_To force this error for JDBMSpace, just change the name (in operation) of the directory where space is saved._

This type of error often appears and causes the `DefaultTimer` singleton to stop working.

I think that the passed `TimerTask` should be protected against such a situation. You would have to add a wrapper to `TimerTask` with error handling with `e.printStackTrace()` logging;

On the other hand, `SpaceError` itself doesn't seem like a good idea to me. See [PMD DoNotExtendJavaLangError](https://pmd.github.io/latest/pmd_rules_java_design.html#donotextendjavalangerror)

> is also a subclass of Error because most applications should not try to catch it.

and [Error javadoc](https://docs.oracle.com/javase/8/docs/api/java/lang/Error.html) and should be converted to unchecked eg. `SpaceException`.

ar:
I think we should never use the DefaultTimer and move to private scheduled executors. Thanks for raising this issue.

PR #324, opened 2020-06-26
Remove deprecated Vector and sendMessageTrailler()

demsey:
Most of changes concentrate on remove deprecated `Vector` also `BaseChannel.sendMessageTrailler()` method

15fb12c introduce small API change - signature of `VErrorParser.getVErrors()` method. It now return the `List<ISOVError>` instead of the raw `Vector`.  

fef8612 partially fixs mentoined in #311 problem with `JDBMSpace`

0f17fb8 contains `ISOServer` refactoring _(use java streams api on channels member)_
30f64c4 contains `BaseChannel`  cleanup


jpos/src/main/java/org/jpos/iso/ISOServer.java
```diff
@@ -654,31 +654,23 @@ public ISOChannel getLastConnectedISOChannel () {
...
-                    sb.append (' ');
-                }
-                sb.append (entry.getKey());
-            }
-        }
-        return sb.toString();
+        return channels.entrySet().stream()
+                .filter(ent -> !LAST.equals(ent.getKey()))
+                .filter(ent -> ent.getValue().get() != null)
+                .filter(ent -> ent.getValue().get().isConnected())
```
ar, line 669: Very rare to occur, but in theory, I think we can have a race condition between line 668 and 669, because value is a WeakReference.

jpos/src/main/java/org/jpos/iso/ISOServer.java
```diff
@@ -700,22 +692,21 @@ public String getCountersAsString () {
...
-            ISOChannel c = (ISOChannel) ref.get ();
-            if (c != null && !LAST.equals (entry.getKey()) && c.isConnected()) {
-                cnt[2]++;
-                if (c instanceof BaseChannel) {
-                    int[] cc = ((BaseChannel)c).getCounters();
-                    cnt[0] += cc[ISOChannel.RX];
-                    cnt[1] += cc[ISOChannel.TX];
+        channels.entrySet().stream()
+            .filter(ent -> !LAST.equals(ent.getKey()))
+            .filter(ent -> ent.getValue().get().isConnected())
```
ar, line 698: Remember `ent.getValue().get()` could be null, and the call to `isConnected()` can cause an NPE.

jpos/src/main/java/org/jpos/iso/ISOServer.java
```diff
@@ -758,27 +749,24 @@ public String getCountersAsString (String isoChannelName) {
...
-                sb.append (Integer.toString (cc[ISOChannel.TX]));
-                sb.append (", last=");
-                sb.append (Long.toString(lastTxn));
-                p.println (sb.toString());
-            }
-        }
+        channels.entrySet().stream()
+            .filter(ent -> !LAST.equals(ent.getKey()))
+            .filter(ent -> ent.getValue().get() instanceof BaseChannel)
+            .filter(ent -> ent.getValue().get().isConnected())
```
ar, line 756: Again, two calls to `WeakReference.get` could return different values.

PR #325, opened 2020-06-30
Use gradle to create q2 start scripts

jameshilliard:
This should make it possible to programmatically modify various aspects of the q2 script at install time.


ar:
Nice contribution, but we want to keep existing startup scripts for now. Will keep this PR around as a reference.

PR #326, opened 2020-07-09
Improve SSL Channels docs

aVolpe:
The current documentation of the [pro guide](http://jpos.org/doc/proguide-draft.pdf) doesn't mention the `server-socket-factory` that is needed to configure a SSL server socket for the QServer.


ar:
Thank you for the contribution to the doc. Updated [proguide here](http://jpos.org/doc/proguide.pdf).

PR #327, opened 2020-08-03
Improve ssh authorized_keys permissions check.

jameshilliard:
The current whitelist approach is overly strict and prevents valid permissions for authorized_keys from being used. Instead we should use a permissions blacklist implementation from openssh.

Use reflection for com.sun.security.auth.module.UnixSystem to avoid compilation failures on Windows.

Fall back to previous behavior if unable to determine the current uid.


ar:
Thank you for this PR and for the follow-up #328 that I've just merged.

PR #334, opened 2020-09-16
Isolate tests using JESpace to their own folders.

jameshilliard:
Lets see if this lets us reliably run tests without `--no-parallel`.


jameshilliard:
I'm somewhat suspecting this could fix the issue I was trying to fix in #231 and #232 as well.

PR #342, opened 2020-09-24
Refactor Q2 to use NIO API for file operations.

jameshilliard:


ar:
For a change like this, the JEP template comes to mind - https://openjdk.java.net/jeps/2

What's the motivation for this PR? what are the goals? what problem are you trying to solve?

barspi:
I'm not sure it's wise to change a **public** API (and interface) that has been working for years and may break things
In particular, the jPOS-EE `FreemarkerDecorationProvider` depends on the old `File` signatures for `ConfigDecorationProvider` 

Also. the behavior and stability of the `Q2` class is **central to everything**, so these changes shouldn't be taken lightly

jameshilliard:
> What's the motivation for this PR? what are the goals? what problem are you trying to solve?

Well I initially started poking around at Q2's file handling code while trying to debug a sporadic test failure issue I was occasionally seeing that seemed to be related to a deploy directory file conflict. I noticed that the public interface exposing legacy `io` file handling is fairly minimal and easy enough to shim so it seemed like a suitable candidate for refactoring to use the modern `nio` API.

While poking around at Q2 it looked like it also might be a decent integration point for some dynamic deployment configuration management tooling as well that I'm working on, I didn't want to have to integrate the legacy blocking `io` file handling with that since there could be performance implications so I first wanted to convert the internal implementation to use `nio`.

jameshilliard:
> I'm not sure it's wise to change a public API (and interface) that has been working for years and risk breaking things.

It wasn't my intention to break the existing public API, I was trying to modernize the internal implementation and then extend the public API with `nio` in a backwards compatible way. Did I break something accidentally?

barspi:
I didn't go deep into the code, but just quickly identified code **external to the jPOS project** such as the `FreemarkerDecorationProvider` in project jPOS-EE.

jameshilliard:
> Granted, we can change that one too, but who knows what else is out there in the wild?

That's why I was trying to make sure that the public interface was still fully backwards compatible.

jameshilliard:
> I didn't go deep into the code, but just quickly identified code external to the jPOS project such as the FreemarkerDecorationProvider in project jPOS-EE.

Oh, I think I see what needs to be fixed there.

jameshilliard:
I reverted the changes to `ConfigDecorationProvider` so it should hopefully now be backwards compatible.

jameshilliard:
One other major reason for this is that the legacy `io` file handling is difficult to use in a thread-safe manner, so I wouldn't want to hook a deployment configuration automation interface into Q2 without the methods being somewhat easy to use in a thread-safe manner.

jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -810,22 +821,22 @@ private void deployBundle (File bundle, boolean encrypt)
         }
     }
     public void deployElement (Element e, String fileName, boolean encrypt, boolean isTransient)
-        throws ISOException, IOException, GeneralSecurityException
+        throws IOException, GeneralSecurityException
```
vishnupillai, line 824: Might be better to leave the ISOException here. It's an API method and subclasses might have over-ridden this method.

jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -83,7 +83,7 @@
  * @author <a href="mailto:vsalaman@vmantek.com">Victor Salaman</a>
  */
 @SuppressWarnings("unchecked")
-public class Q2 implements FileFilter, Runnable {
+public class Q2 implements DirectoryStream.Filter<Path>, Runnable {
```
vishnupillai, line 86: Is it possible to leave the FileFilter interface? I don't think it's a big deal but might help minimize impact for anyone updating to this version

vishnupillai:
nice one @jameshilliard I think the changes are good - added 2 comments above

ar:
We don't have a performance problem deploying QBeans. Nobody is going to deploy 1000 QBeans per second in the near future. I'm closing this PR, specially since it will break existing code.

jameshilliard:
> We don't have a performance problem deploying QBeans.

Yeah, it's mostly for thread safety not performance.

> I'm closing this PR, specially since it will break existing code.

Ok, I'll rework this to fix that.

PR #344, opened 2020-09-28
Q2: Add getXmlConfigMap() method for fetching configs.

jameshilliard:
This provides a method to retrieve the xml config data programmatically in a map format, for there it can easily be converted to json.
For example:
```json
{
    "99_sysmon": {"sysmon": {
        "property": {"metrics-dir": "log"},
        "attr": {
            "SleepTime": 3600000,
            "DetailRequired": true
        }
    }},
    "00_logger": {"logger": {
        "property": {"redirect": "stdout, stderr"},
        "log-listener": [
            {"log-listener": {"class": "org.jpos.util.SimpleLogListener"}},
            {"log-listener": {
                "property": {
                    "compression-format": "gzip",
                    "window": "86400",
                    "suffix": ".log",
                    "date-format": "-yyyy-MM-dd-HH",
                    "prefix": "log/q2"
                },
                "class": "org.jpos.util.DailyLogListener"
            }}
        ],
        "class": "org.jpos.q2.qbean.LoggerAdaptor"
    }}
}
```

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/Q2.java
+++ b/jpos/src/main/java/org/jpos/q2/Q2.java
@@ -31,6 +31,8 @@ import org.jdom2.JDOMException;
 import org.jdom2.input.SAXBuilder;
 import org.jdom2.output.Format;
 import org.jdom2.output.XMLOutputter;
+import org.jpos.core.Configuration;
+import org.jpos.core.ConfigurationException;
 import org.jpos.core.Environment;
 import org.jpos.iso.ISOException;
 import org.jpos.iso.ISOUtil;
@@ -52,12 +54,7 @@ import org.osgi.framework.launch.FrameworkFactory;
 import org.xml.sax.SAXException;
 import javax.crypto.Cipher;
 import javax.crypto.spec.SecretKeySpec;
-import javax.management.InstanceAlreadyExistsException;
-import javax.management.InstanceNotFoundException;
-import javax.management.MBeanServer;
-import javax.management.MBeanServerFactory;
-import javax.management.ObjectInstance;
-import javax.management.ObjectName;
+import javax.management.*;
 import java.io.*;
 import java.lang.management.ManagementFactory;
 import java.nio.file.FileSystem;
@@ -73,6 +70,7 @@ import java.time.Instant;
 import java.util.*;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Stream;
 
 import static java.util.ResourceBundle.getBundle;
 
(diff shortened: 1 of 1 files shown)


jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -360,6 +360,84 @@ public static Q2 getQ2(long timeout) {
...
+        String className = e.getAttributeValue("class");
+        if (className != null) {
+            map.put("class", className);
+        }
+        List<Element> children = e.getChildren();
+        Map<String, Object> props = null;
+        Map<String, Object> attr = null;
+        for(Element child: children){
+            String key = child.getName();
+            if (key == "property") {
```
ar, line 392: Shouldn't you use `"property".equals(key)` here ?

jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -360,6 +360,84 @@ public static Q2 getQ2(long timeout) {
...
+                if (props == null) {
+                    props = new HashMap<>();
+                }
+                props.put(child.getAttributeValue("name"), child.getAttributeValue("value"));
+            } else if (key == "attr") {
```
ar, line 397: Shouldn't you use `"attr".equals(key)` here ?

jpos/src/main/java/org/jpos/q2/Q2.java
barspi: I think using `==` is a Kotlin idiom ;-)

jameshilliard: > I think using `==` is a Kotlin idiom ;-)

Hmm, so that looks to be correct, in Koltin "The equals method returns the same result as the “==” operator" while in Java it seems it doesn't, I find that Java behavior to be strange but jPOS is really the first Java based project I've worked on so I'm less familiar with Java semantics.

The programming language I have the most experience in is python however where `==` is a standard value equality check which seems to be fairly common across most programming languages in general.

I regularly work with many different programming languages and usually just expect to catch these sort of issues during testing/review since it's difficult to remember all the minor syntax differences when jumping around between projects.

jameshilliard: Pushed a fix changing this to `"property".equals(key)`.

jpos/src/main/java/org/jpos/q2/Q2.java
jameshilliard: Pushed a fix changing this to` "attr".equals(key)`.

jameshilliard:
Refactored this a bit more to better use existing methods for parsing element trees into maps. Example json dump updated.

ar:
It is not the Q2 business to expose QBean configuration, that may contain sensitive data.
On the contrary, its business is to protect it.

PR #345, opened 2020-09-28
Add method to get Element Properties without SimpleConfiguration.

jameshilliard:


ar:
The responsibility of the SimpleConfiguration is to implement a Configuration object.

The fact that SimpleConfiguration internally use Properties is an implementation fact, that may change in the future.

PR #346, opened 2020-09-28
Add method to get Element Map.

jameshilliard:


ar:
The responsibility of the `SimpleConfigurationFactory` is to create a `Configuration` object.

The fact that SimpleConfiguration internally use Properties is an implementation fact, that may change in the future.

PR #347, opened 2020-10-01
Experimental jolokia based Q2 config element json converter.

jameshilliard:
Potential alternative to #344 based on jolokia. I'm not entirely sure how it would get integrated but jolokia seems to allow for [json based mbeans](https://jolokia.org/reference/html/jmx.html#json-mbean) as well.

Example config dump:
```json
{
    "99_sysmon": {
        "name": "sysmon",
        "property": {"metrics-dir": "log"},
        "attr": {
            "SleepTime": 3600000,
            "DetailRequired": true
        }
    },
    "00_logger": {
        "name": "logger",
        "property": {"redirect": "stdout, stderr"},
        "objectName": "org.jpos.q2.qbean.LoggerAdaptor",
        "hasChildNodes": [
            {
                "name": "log-listener",
                "objectName": "org.jpos.util.SimpleLogListener"
            },
            {
                "name": "log-listener",
                "property": {
                    "compression-format": "gzip",
                    "window": "86400",
                    "suffix": ".log",
                    "date-format": "-yyyy-MM-dd-HH",
                    "prefix": "log/q2"
                },
                "objectName": "org.jpos.util.DailyLogListener"
            }
        ]
    }
}
```

### DIFF
--- a/jpos/build.gradle
+++ b/jpos/build.gradle
@@ -25,8 +25,10 @@ dependencies {
 
     implementation libraries.osgi_core
     implementation libraries.javassist
+    implementation libraries.jolokia_jmx
     implementation libraries.yaml;
     implementation libraries.jdbm
+    implementation libraries.json
     implementation libraries.sleepycat_je
     implementation libraries.sshd
     implementation libraries.eddsa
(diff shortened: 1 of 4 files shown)


ar:
I don't think these kind of experiments, that add dependencies to jPOS, belong here. This can be done as part of your internal code, and once battle tested in production, can be considered as a jPOS-EE add-on module.

PR #352, opened 2020-10-08
BERTLVPackager can't pack non EMVStandardTagType

### LINKED ISSUE #349: BERTLVPackager can't pack non EMVStandardTagType tags
Seems this issue has been introduced after a #315 fix.

We have a list of tags which doesn't belong to a standard EMV tags and have an UnknownTagNumberException in the "2.1.5-20201001" version, while the version "2.1.4" doesn't have a such issue.

 protected byte[] packValue(String tagNameHex, final ISOComponent c) throws ISOException,
            UnknownTagNumberException {
        final int tagNumber = Integer.parseInt(tagNameHex, 16);
        final TLVDataFormat dataFormat = getTagFormatMapper().getFormat(tagNumber);
        String tagValue;
        byte[] packedValue;

        if (c.getComposite() == null) {
            if (c.getValue() instanceof String) {
                tagValue = (String) c.getValue();
                EMVStandardTagType tagType = EMVStandardTagType.forHexCode(tagNameHex);
                int length = Math.max(tagValue.length(), tagType.getDataLength().getMinLength());
                switch (dataFormat) {

obviously custom tag doesn't belong to the EMVStandardTagType and this component throws an exception.

ar:
attempt to fix #349

### DIFF
--- a/jpos/src/main/java/org/jpos/tlv/packager/bertlv/BERTLVPackager.java
+++ b/jpos/src/main/java/org/jpos/tlv/packager/bertlv/BERTLVPackager.java
@@ -338,8 +338,14 @@ public abstract class BERTLVPackager extends GenericPackager {
         if (c.getComposite() == null) {
             if (c.getValue() instanceof String) {
                 tagValue = (String) c.getValue();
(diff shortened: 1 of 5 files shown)


dark1st:
packing is working well for me now.

can you merge this request?

PR #366, opened 2020-12-01
Important key block fixes

demsey:
The `SecureKeyBlockBuilder` cannot properly parse keys formatted as key blok because the length of `MAC` was misinterpreted.


ar (approved):

demsey:
This check failed due to a typo in 1728d2c in #367

ar (approved):

PR #368, opened 2020-12-03
Add github workflow and fix some unreliable tests.

jameshilliard:


ar:
This PR, while good, is a mega PR that touches totally unrelated stuff. The title is misleading, it says it fixes some unreliable tests, but then it changes core components such as TSpace, JESpace, JDBMSpace.

Please send isolated PRs so we can test separately. We are close to release 2.1.5 and while we can accept the test related ones, and the github actions one, this is not a good time to change core components.

jpos/src/test/java/org/jpos/iso/SslChannelIntegrationTest.java
```diff
@@ -65,7 +65,14 @@ public void serverSideDisconnect() throws Exception {
 
         XMLChannel clientChannel = newClientChannel();
 
-        clientChannel.connect();
+        while (!clientChannel.isConnected()) {
```
ar, line 68: This can loop forever and could hog the tests. I'd add a counter, and a small delay within the loop.

jameshilliard:
> We are close to release 2.1.5 and while we can accept the test related ones, and the github actions one, this is not a good time to change core components.

Well the timer changes were essentially a prerequisite for the rest from what things were looking like.

PR #385, opened 2020-12-17
Lots of refactoring to improve clock/test reliability.

jameshilliard:
I'll clean up and split off parts of this in a little to separate pull requests but I think I finally got all github action tests working somewhat reliably.

Lots of OSX CI issues worked around:

- Relax timing accuracy for various tests due to major timer+scheduling accuracy issues in the OSX test runners.
- Disabled jdbm and je space transaction sync in tests due to extremely slow/unpredictable disk sync latency in github's OSX test runners.

Some other refactoring:

- Guard against accidentally calling Object.wait with `0` which would make the timeout infinite.
- Add a high accuracy `NanoClock` `Clock` class and use it when calling `Instant.now()`.
- Eliminate remaining usage of `System.currentTimeMillis()`.


jpos/src/main/java/org/jpos/iso/ISODate.java
```diff
@@ -131,7 +134,7 @@ public static Date parseDateTime(String s, TimeZone timeZone) {
      * @return Date
      */
     public static Date parseISODate (String d) {
-        return parseISODate (d, System.currentTimeMillis());
+        return parseISODate (d, Instant.now(NanoClock.systemUTC()).toEpochMilli());
```
ar, line 137: We are talking about YEAR accuracy here, makes no sense to use nanoseconds. This will just slow down a method heavily used in most messages going through the system.

jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -59,11 +60,23 @@
...
+     * @param sync sync transactions to disk
+     */
+    protected JDBMSpace (String name, String filename, boolean sync) {
         super();
         this.name = name;
         try {
             Properties props = new Properties();
             props.put (RecordManagerOptions.CACHE_SIZE, "512");
+            if (!sync)
+                props.put (RecordManagerOptions.DISABLE_TRANSACTIONS, "true");
```
ar, line 79: The JDBMSpace rely on having transactions. We don't want users to disable that feature, for their protection. If what we want is just letting the tests pass, we can just go ahead and disable the tests altogether.

jpos/src/main/java/org/jpos/space/JESpace.java
```diff
@@ -71,14 +71,20 @@
...
+
+    public JESpace(String name, String params, boolean sync) throws SpaceError {
         super();
         try {
             EnvironmentConfig envConfig = new EnvironmentConfig();
             StoreConfig storeConfig = new StoreConfig();
             String[] p = ISOUtil.commaDecode(params);
             String path = p[0];
             envConfig.setAllowCreate (true);
+            if (!sync)
```
ar, line 86: See my comment in JDBMSpace. Having around an easy to misconfigure JESpace due to no syncs is a recipe for disaster.

PR #388, opened 2020-12-17
Replace remaining System.currentTimeMillis() usage with Instant.now()

jameshilliard:


ar:
We don't want to incur in additional object creation in inner parts of jPOS to get a nano-level accuracy that we don't need.

jameshilliard:
I've entirely removed `NanoClock` already.

alcarraz:
> We don't want to incur in additional object creation in inner parts of jPOS to get a nano-level accuracy that we don't need.

I just found [this interesting related thing](https://stackoverflow.com/a/61947549/3444205) that I haven't noticed before, maybe we should use [System.nanoTime()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--) for elapsed time measurements or timeout calculation.

Not because of the precision  but for the accuracy in timelapse, if I understand correctly `System.currentTimeMillis` could eventually go backwards?:


> I want to add that System.nanoTime() is less about precision but more about accuracy.
> 
> System.currentTimeMillis() is based on the system clock, which is, most of the time, based on a quartz clock inside a computer. It is not accurate and it drifts. (VM is even worse since you don't have a physical clock and have to sync with the host) When your computer syncs this quartz clock with a global clock, you might even observe your clock jumps backward/forward because your local clock is too fast or slow.
> 
> On the other hand, System.nanoTime() is based on a monotonic clock. This clock has nothing to do with the actual time we humans speak. It only moves forward at a constant pace. It does not drift like the quartz clock and there is no sync required. This is why it is perfect for measuring elapses.

To be clear I'm not proposing to change the existing implementations, just wanted to bring it to the table for future reference.

jameshilliard:
From my understanding `Instant.now()` is generally nanoTime based, and using the monotonic API's was one reason for this change.

jameshilliard:

PR #391, opened 2020-12-18
Replace remaining System.currentTimeMillis() usage with Instant.now()

jameshilliard:
Alternative to #388 that avoid changing any logic.

### DIFF
--- a/jpos/src/main/java/org/jpos/bsh/BSHLogListener.java
+++ b/jpos/src/main/java/org/jpos/bsh/BSHLogListener.java
@@ -29,6 +29,7 @@ import java.io.FileReader;
 import java.io.IOException;
 import java.io.Reader;
 import java.io.StringReader;
+import java.time.Instant;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -148,8 +149,8 @@ public class BSHLogListener implements org.jpos.util.LogListener, org.jpos.core.
                         }
                     }else{
                         if(info == null) scripts.put(sources[i], info=new ScriptInfo());
-                        if(System.currentTimeMillis() > info.getLastCheck() + cfg.getLong("reload")){
-                            info.setLastCheck(System.currentTimeMillis());
+                        if(Instant.now().toEpochMilli() > info.getLastCheck() + cfg.getLong("reload")){
+                            info.setLastCheck(Instant.now().toEpochMilli());
                             if(f.exists() && f.canRead() && f.isFile()){
                                 if(info.getLastModified() != f.lastModified()) {
                                     info.setLastModified(f.lastModified());
--- a/jpos/src/main/java/org/jpos/iso/ISODate.java
+++ b/jpos/src/main/java/org/jpos/iso/ISODate.java
@@ -20,6 +20,7 @@ package org.jpos.iso;
 
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
+import java.time.Instant;
 import java.util.*;
 
 /**
@@ -131,7 +132,7 @@ public class ISODate {
      * @return Date
      */
     public static Date parseISODate (String d) {
-        return parseISODate (d, System.currentTimeMillis());
+        return parseISODate (d, Instant.now().toEpochMilli());
     }
 
     /**
@@ -143,7 +144,7 @@ public class ISODate {
      * @return Date
      */
     public static Date parseISODate (String d, TimeZone timeZone) {
-        return parseISODate (d, System.currentTimeMillis(), timeZone);
+        return parseISODate (d, Instant.now().toEpochMilli(), timeZone);
     }
 
     /**
--- a/jpos/src/main/java/org/jpos/iso/ISOServer.java
+++ b/jpos/src/main/java/org/jpos/iso/ISOServer.java
@@ -29,6 +29,7 @@ import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
 import java.net.UnknownHostException;
+import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EventObject;
@@ -342,7 +343,7 @@ public class ISOServer extends Observable
                 for (;;) {
                     try {
                         ISOMsg m = channel.receive();
-                        lastTxn = System.currentTimeMillis();
+                        lastTxn = Instant.now().toEpochMilli();
                         Iterator iter = listeners.iterator();
                         while (iter.hasNext()) {
                             if (((ISORequestListener)iter.next()).process
@@ -695,7 +696,7 @@ public class ISOServer extends Observable
         sb.append (lastTxn);
         if (lastTxn > 0) {
             sb.append (", idle=");
-            sb.append(System.currentTimeMillis() - lastTxn);
+            sb.append(Instant.now().toEpochMilli() - lastTxn);
             sb.append ("ms");
         }
         return sb.toString();
@@ -742,7 +743,7 @@ public class ISOServer extends Observable
     }
     @Override
     public long getIdleTimeInMillis() {
-        return lastTxn > 0L ? System.currentTimeMillis() - lastTxn : -1L;
+        return lastTxn > 0L ? Instant.now().toEpochMilli() - lastTxn : -1L;
     }
 
 
--- a/jpos/src/main/java/org/jpos/q2/Q2.java
+++ b/jpos/src/main/java/org/jpos/q2/Q2.java
@@ -985,7 +985,7 @@ public class Q2 implements FileFilter, Runnable {
         }
     }
     private void logVersion () {
-        long now = System.currentTimeMillis();
+        long now = Instant.now().toEpochMilli();
         if (now - lastVersionLog > 86400000L) {
             LogEvent evt = getLog().createLogEvent("version");
             evt.addMessage(getVersionString());
--- a/jpos/src/main/java/org/jpos/q2/cli/TAIL.java
+++ b/jpos/src/main/java/org/jpos/q2/cli/TAIL.java
@@ -26,6 +26,7 @@ import org.jpos.util.Logger;
 import org.jpos.util.NameRegistrar;
 
 import java.io.PrintStream;
+import java.time.Instant;
 import java.util.Date;
 import java.util.Iterator;
 import java.util.Map;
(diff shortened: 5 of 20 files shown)


ar:
In Java8, after unnecessary Object creation, this code ends up calling `System.currentTimeMillis()`.

```java
   @Override
        public long millis() {
            return System.currentTimeMillis();
        }
        @Override
        public Instant instant() {
            return Instant.ofEpochMilli(millis());
        }
```

jameshilliard:
Looks like the precision increase might have only made it into Java 9 and newer, see [here](https://github.com/AdoptOpenJDK/openjdk-jdk/commit/b5f5424c83e28240a3eb280251ffe604a294ac20).

jameshilliard:
We could use [`NanoClock`](https://github.com/jameshilliard/jPOS/blob/nanoclock/jpos/src/main/java/org/jpos/util/NanoClock.java) for Java 8 potentially.

ar:
In which specific method do you believe we need nano resolution?

alcarraz:
Hi, I believe I already pointed this out in a comment in a previous PR, maybe what we want is [System.nanoTime()](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--), it's not about nanosecond resolution but more about accuracy between two different moments in time. i.e. when we want to calculate elapsed times.

As I understand there is no warranty that leap millies are inserted in `System.currentTimeMillis()` because it depends on the underlying OS current date, while `nanoTime` uses the internal clock to increase constant ticks.

I'm not saying that is worth that other change, that would not involve object creation but an operation to divide over `1000` here or there, which can be solved by multiplying timeouts by `1000` at config time.

What I am saying is that if we want better accuracy we don't need object creation, the usage of `System.nanoTime()`  will suffice, and again is not that it will provide nanosecond resolution but it will provide better elapsed time consistency

jameshilliard:
> In which specific method do you believe we need nano resolution?

As @alcarraz mentioned it was mostly relating getting monotonic behavior, my original implementation used [this](https://github.com/jameshilliard/jPOS/blob/nanoclock/jpos/src/main/java/org/jpos/util/NanoClock.java) to ensure we used `System.nanoTime()` for monotonic elapsed time behavior, looking at it again this might be required for monotonic behavior on newer Java versions as well so thinking I need to restore that.

alcarraz:
> > In which specific method do you believe we need nano resolution?
> 
> As @alcarraz mentioned it was mostly relating getting monotonic behavior, my original implementation used [this](https://github.com/jameshilliard/jPOS/blob/nanoclock/jpos/src/main/java/org/jpos/util/NanoClock.java) to ensure we used `System.nanoTime()` for monotonic elapsed time behavior, looking at it again this might be required for monotonic behavior on newer Java versions as well so thinking I need to restore that.

That's correct @jameshilliard, but what I say is that we don't need new object creation to achieve that, which is what @ar wants to avoid for frequent operations. Just call `System.nanoTime()` instead of creating an `Instant` from a clock just to end calling `System.nanoTime()`

jameshilliard:
> Just call `System.nanoTime()` instead of creating an `Instant` from a clock just to end calling `System.nanoTime()`

I can do this in some places, but `System.nanoTime()` is not really a direct substitution as it uses a runtime based epoch from my understanding, so it works for elapsed time but not for use cases where we need an actual epoch timestamp(we can use `NanoClock` with `Instant.now()` for those instances however).

alcarraz:
from [`System.nanoTime`](https://docs.oracle.com/javase/8/docs/api/java/lang/System.html#nanoTime--)


> nanoTime

> `public static long nanoTime()`
> Returns the current value of the running Java Virtual Machine's high-resolution time source, in nanoseconds.
> This method can only be used to measure elapsed time and is not related to any other notion of system or wall-clock time. The value returned represents nanoseconds since some fixed but arbitrary origin time (perhaps in the future, so values may be negative). The same origin is used by all invocations of this method in an instance of a Java virtual machine; other virtual machine instances are likely to use a different origin.
> 
> This method provides nanosecond precision, but not necessarily nanosecond resolution (that is, how frequently the value changes) - no guarantees are made except that the resolution is at least as good as that of currentTimeMillis().
> 
> Differences in successive calls that span greater than approximately 292 years (263 nanoseconds) will not correctly compute elapsed time due to numerical overflow.
> 
> The values returned by this method become meaningful only when the difference between two such values, obtained within the same instance of a Java virtual machine, is computed.
> 
> For example, to measure how long some code takes to execute:
> 
>  
```
long startTime = System.nanoTime();
// ... the code being measured ...
long estimatedTime = System.nanoTime() - startTime;
```
> To compare two nanoTime values
> 
>  
```
long t0 = System.nanoTime();
...
long t1 = System.nanoTime();
```
> one should use `t1 - t0 < 0`, not `t1 < t0`, because of the possibility of numerical overflow.
> Returns:
> the current value of the running Java Virtual Machine's high-resolution time source, in nanoseconds
> Since:
> 1.5

alcarraz:
> > Just call `System.nanoTime()` instead of creating an `Instant` from a clock just to end calling `System.nanoTime()`
> 
> I can do this in some places, but `System.nanoTime()` is not really a direct substitution as it uses a runtime based epoch from my understanding, so it works for elapsed time but not for use cases where we need an actual epoch timestamp(we can use `NanoClock` with `Instant.now()` for those instances however).

Yes but there is when @ar's point holds, why do we need nano accuracy for absolute times?

jameshilliard:
> Yes but there is when @ar's point holds, why do we need nano accuracy for absolute times?

We don't, but we do want monotonic behavior for absolute times as well.

alcarraz:
> > Yes but there is when @ar's point holds, why do we need nano accuracy for absolute times?
> 
> We don't, but we do want monotonic behavior for absolute times as well.

can you show a use case where that is critical?

jameshilliard:
> can you show a use case where that is critical?

Look at the space timeout code. It's epoch based but expects monotonic behavior from my understanding.

alcarraz:
> Look at the space timeout code. It's epoch based but expects monotonic behavior from my understanding.

Ok then you just need to make a PR for those classes where it matters and explaining the why, and not replace every single line where we make a call to `system.currentTimeMillis()` even though I don't see how `NanoClock` solves any issue since it calculates an initial offset that it maintains forever, so, if `nanoTime()` is not monotonic, the result of `NanoClock` won't be either unless I'm missing something.

jameshilliard:
> NanoClock solves any issue since it calculates an initial offset that it maintains forever, so, if nanoTime() is not monotonic, the result of NanoClock won't be either unless I'm missing something.

Well the idea is that by only using the non-monotonic `system.currentTimeMillis()` once to compute the initial offset you ensure that all derived `nanoTime()` offset calculations are monotonic in relation to the initial offset computation, at least until you shut it down.

alcarraz:
> Well the idea is that by only using the non-monotonic `system.currentTimeMillis()` once to compute the initial offset you ensure that all derived `nanoTime()` offset calculations are monotonic in relation to the initial offset computation, at least until you shut it down.

Ok I see, then it's only relevant to persistent or distributed spaces, you could make a PR only target those places, but everywhere else is not needed and in the other places for elapsed times you can just rely on `System.nanoTime` to avoid unnecessary object creations

jameshilliard:
> Ok I see, then it's only relevant to persistent or distributed spaces, you could make a PR only target those places, but everywhere else is not needed and in the other places for elapsed times you can just rely on `System.nanoTime` to avoid unnecessary object creations

Yeah, I'll refactor this to eliminate the object creation, I hadn't done that originally since it was a bit harder to read than using `Instant.now()` with `NanoClock`.

jameshilliard:
> then it's only relevant to persistent or distributed spaces, you could make a PR only target those places, but everywhere else is not needed

Seems a good bit of other code expects monotonic behavior and also requires epoch's, did an initial refactoring of code using `Instant.now()` to use `System.nanoTime()` in #393. I was able to use pure elapsed time durations without the clock for some of the timeouts at least.

ar:
We don't need nanoPrecision. We don't need monotonic behavior even in space operations, the space timeouts are usually measured in seconds. I don't see a reason to apply this massive PR. The fact that `System.nanoTime()` can give us negative value raises many flags with this kind of search & replace PRs.

You are solving a problem that we don't have. First specify the problem, in an issue, and we can evaluate a work around.

jameshilliard:
> We don't need monotonic behavior even in space operations, the space timeouts are usually measured in seconds.

I don't think this is true, the space timeout code to me appears to expect monotonic behavior for the time delta computations to work as expected.

> The fact that System.nanoTime() can give us negative value raises many flags with this kind of search & replace PRs.

In #393 I'm wrapping `System.nanoTime()` in a `Duration` for this reason and separate `Duration`'s from `Instant`'s to make it easier to verify that nothing from a pure duration/elapsed time computation is not accidentally used in place of a unix timestamp as `Duration` doesn't provide a method to get a UTC timestamp like `Instant` does.

> You are solving a problem that we don't have. First specify the problem, in an issue, and we can evaluate a work around.

This isn't an issue that would get hit all that often in production for jpos, but I'm pretty sure it is a real issue. For some background I've hit this type of clock jump issue before most often when working with embedded systems without battery backed hardware clocks, essentially during startup the device would sync the clock from NTP and any running apps would have their computations get messed up by the large clock jump if the application was not using monotonic clock based computations for elapsed time calculations. This is much more uncommon for normal servers as they have battery backed clocks but it is still a possibility. For example you could hit this issue in production if someone installs a ntp client on a server while jpos is running as the ntp clock sync could cause a relatively large non-monotonic clock jump.

I'm basically trying to guard against a class of edge cases that can be hit if the clock changes in between start timestamp and end timestamp. The change to nanoclock also seems to increase timeout related test reliability on some platforms on github actions as the non-monotonic clocks on OSX/Windows seem to be less reliable.

jameshilliard:
Here's [an article](https://itnext.io/as-a-software-developer-why-should-you-care-about-the-monotonic-clock-7d9c8533595c) on the main issue I'm referring to.

ar:
Check this: https://serverfault.com/questions/94683/will-ntp-drift-the-clock-backwards

While it can certainly go backwards if manually forced, that's a very rare thing to happen.

That said, thank you for your PR, we'll keep it in mind.

jameshilliard:
> While it can certainly go backwards if manually forced, that's a very rare thing to happen.

I think it's a forward jump that would be the most likely issue to come up, which appears could cause premature space timeouts and other messed up calculations along those lines.

jameshilliard:
> Check this: https://serverfault.com/questions/94683/will-ntp-drift-the-clock-backwards

FYI there are multiple different ntp daemons in use these days on Linux systems, and I'm not sure they all have that behavior, the other one I'm most familiar with that's widely used is [systemd-timesyncd](https://www.freedesktop.org/software/systemd/man/systemd-timesyncd.service.html), and of course there are the non-Linux platforms as well. I think one other potential cause for clock jumps is having one's configured network time servers being out of sync as well.

jameshilliard:

PR #393, opened 2020-12-29
Refactor to use System.nanoTime() based clock/timeouts.

jameshilliard:
See discussion in #391. Purpose is to ensure a monotonic clock is used for duration based computations.

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/Q2.java
+++ b/jpos/src/main/java/org/jpos/q2/Q2.java
@@ -116,7 +116,7 @@ public class Q2 implements FileFilter, Runnable {
     private String[] args;
     private boolean hasSystemLogger;
     private boolean exit;
-    private Instant startTime;
+    private Duration startTime;
     private CLI cli;
     private boolean recursive;
     private ConfigDecorationProvider decorator=null;
@@ -142,7 +142,7 @@ public class Q2 implements FileFilter, Runnable {
     public Q2 (String[] args, BundleContext bundleContext) {
         super();
         this.args = args;
-        startTime = Instant.now();
+        startTime = Duration.ofNanos(System.nanoTime());
         instanceId = UUID.randomUUID();
         parseCmdLine (args);
         libDir     = new File (deployDir, "lib");
@@ -673,7 +673,7 @@ public class Q2 implements FileFilter, Runnable {
         return server;
     }
     public long getUptime() {
-        return Duration.between(startTime, Instant.now()).toMillis();
+        return Duration.ofNanos(System.nanoTime()).minus(startTime).toMillis();
     }
     public void displayVersion () {
         System.out.println(getVersionString());
--- a/jpos/src/main/java/org/jpos/q2/cli/TZCHECK.java
+++ b/jpos/src/main/java/org/jpos/q2/cli/TZCHECK.java
@@ -20,6 +20,7 @@ package org.jpos.q2.cli;
 
 import org.jpos.q2.CLICommand;
 import org.jpos.q2.CLIContext;
+import org.jpos.util.NanoClock;
 
 import java.time.Instant;
 import java.time.ZoneId;
@@ -35,7 +36,7 @@ public class TZCHECK implements CLICommand
     public void exec(CLIContext cli, String[] args) throws Exception
     {
         ZoneId zi = ZoneId.systemDefault();
-        Instant i = Instant.now();
+        Instant i = Instant.now(NanoClock.systemUTC());
         cli.println(
             "         Zone ID: " + zi + " (" + zi.getDisplayName(TextStyle.FULL, Locale.getDefault()) + ") "
                 + zi.getRules().getOffset(i)
--- a/jpos/src/main/java/org/jpos/q2/qbean/SystemMonitor.java
+++ b/jpos/src/main/java/org/jpos/q2/qbean/SystemMonitor.java
@@ -36,6 +36,7 @@ import java.lang.management.ThreadMXBean;
 import java.net.InetAddress;
 import java.nio.charset.Charset;
 import java.security.NoSuchAlgorithmException;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.time.format.TextStyle;
@@ -43,7 +44,6 @@ import java.time.zone.ZoneOffsetTransition;
 import java.time.zone.ZoneOffsetTransitionRule;
 import java.util.List;
 import java.util.Locale;
-import java.util.Map;
 
 /**
  * Periodically dumps Thread and memory usage
@@ -55,8 +55,8 @@ import java.util.Map;
 public class SystemMonitor extends QBeanSupport
         implements Runnable, SystemMonitorMBean, Loggeable
 {
-    private long sleepTime = 60 * 60 * 1000;
-    private long delay = 0;
+    private Duration sleepTime = Duration.ofSeconds(60 * 60);
+    private Duration delay = Duration.ofSeconds(0);
     private boolean detailRequired = false;
     private Thread me = null;
     private static final int MB = 1024*1024;
@@ -82,14 +82,14 @@ public class SystemMonitor extends QBeanSupport
     }
 
     public synchronized void setSleepTime(long sleepTime) {
-        this.sleepTime = sleepTime;
+        this.sleepTime = Duration.ofMillis(sleepTime);
         setModified(true);
         if (me != null)
             me.interrupt();
     }
 
     public synchronized long getSleepTime() {
-        return sleepTime;
+        return sleepTime.toMillis();
     }
 
     public synchronized void setDetailRequired(boolean detail) {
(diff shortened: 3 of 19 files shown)


jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -191,7 +192,7 @@ public void out (K key, V value, long timeout) {
                 long recid = recman.insert (value);
 
                 long expiration = timeout == -1 ? Long.MAX_VALUE :
-                        Instant.now().toEpochMilli() + timeout;
+                        Instant.now(NanoClock.systemUTC()).toEpochMilli() + timeout;
```
alcarraz, line 195: I believe here the use of `NanoClock`can be justified

jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -237,7 +238,7 @@ public void push (Object key, Object value, long timeout) {
             synchronized (this) {
                 long recid = recman.insert (value);
                 long expiration = timeout == -1 ? Long.MAX_VALUE :
-                        Instant.now().toEpochMilli() + timeout;
+                        Instant.now(NanoClock.systemUTC()).toEpochMilli() + timeout;
```
alcarraz, line 241: I believe here the use of `NanoClock`can be justified

jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -665,7 +670,7 @@ public Ref (long recid, long expires) {
         }
 
         public boolean isExpired () {
-            return expires < Instant.now().toEpochMilli();
+            return expires < Instant.now(NanoClock.systemUTC()).toEpochMilli();
```
alcarraz, line 673: I believe here the use of `NanoClock`can be justified

jpos/src/main/java/org/jpos/space/TSpace.java
```diff
@@ -18,6 +18,8 @@
 
 package org.jpos.space;
 import org.jpos.util.Loggeable;
+import org.jpos.util.NanoClock;
```
alcarraz, line 21: `TSpace` instances live inside the scope of a single JVM, we don't need `NanoClock` here, it only adds overhead to a really lightweight `Space` implementation. just use `System.nanoTime()`

jpos/src/main/java/org/jpos/transaction/TransactionManager.java
```diff
@@ -1114,7 +1117,7 @@ public static Long getId() {
     private void setThreadName (long id, String method, TransactionParticipant p) {
         Thread.currentThread().setName(
             String.format("%s:%d %s %s %s", getName(), id, method, p.getClass().getName(),
-                LocalDateTime.ofInstant(Instant.now(), ZoneId.systemDefault()))
+                LocalDateTime.ofInstant(Instant.now(NanoClock.systemUTC()), ZoneId.systemDefault()))
```
alcarraz, line 1120: We don't need NanoClock for monotonic during the living of the same instance of the jvm, `System.nanoTime()` already provides that

jpos/src/main/java/org/jpos/util/TPS.java
```diff
@@ -80,7 +80,7 @@ public TPS(boolean autoupdate) {
     public TPS(final long period, boolean autoupdate) {
         super();
         count = new AtomicInteger(0);
-        start = peakWhen = Instant.now();
+        start = peakWhen = Instant.now(NanoClock.systemUTC());
```
alcarraz, line 83: We don't need NanoClock for monotonic during the living of the same instance of the jvm, `System.nanoTime()` already provides that

alcarraz (changes requested):
Hi @jameshilliard I could do through all the code changes, but I pointed out some places where duration are used instead of Instants and places where `NanoClock` is not actually needed and some places where it is or can be justified.

jpos/src/main/java/org/jpos/space/JDBMSpace.java
ar: For persisted objects, the timeouts are usually measured in minutes, hours, or no timeout. Not sure we need nano precision here.

jpos/src/main/java/org/jpos/util/TPS.java
ar: peakWhen just give us an approximate time where the TPS peaked, it's intended to be used by a human to get an idea of the time of day, and eventually correlate it to other graphs. No need for nano precision.

jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -116,7 +116,7 @@
     private String[] args;
     private boolean hasSystemLogger;
     private boolean exit;
-    private Instant startTime;
+    private Duration startTime;
```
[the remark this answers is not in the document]
jameshilliard: Pure uptime calculations using `System.nanoTime()` without any epoch based API's are `Duration`s since `System.nanoTime()` is effectively a JVM runtime `Duration` itself. We only need to use `Instant` when we need an epoch.

jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -322,13 +323,14 @@ public synchronized V in (Object key) {
      */
     public synchronized V in (Object key, long timeout) {
         Object obj;
-        Instant now = Instant.now();
-        long duration;
+        Duration to = Duration.ofMillis(timeout);
+        Duration now = Duration.ofNanos(System.nanoTime());
```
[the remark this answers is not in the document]
jameshilliard: Now uses `System.nanoTime()` which by itself is actually a `Duration` not an `Instant`. We only need to use `Instant` when we need an epoch.

jpos/src/main/java/org/jpos/space/JDBMSpace.java
jameshilliard: Well I'm trying to ensure we have monotonic behavior here as much as possible, hence the use of `NanoClock`.

jpos/src/main/java/org/jpos/q2/Q2.java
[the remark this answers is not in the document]
ar: We use `uptime` to get an idea if the system was booted a few hours ago, a few days ago, or a few months ago.

What's the rationale behind having nano precision here?

jpos/src/main/java/org/jpos/space/TSpace.java
jameshilliard: True, let me try and refactor that.

jpos/src/main/java/org/jpos/transaction/TransactionManager.java
jameshilliard: Well this is epoch based from what I can tell, and `System.nanoTime()` is only a duration by itself.

jpos/src/main/java/org/jpos/util/TPS.java
jameshilliard: Well I'm trying to ensure the TPS computations are fully monotonic here, hence the use of `NanoClock`.

jpos/src/main/java/org/jpos/space/JDBMSpace.java
```diff
@@ -322,13 +323,14 @@ public synchronized V in (Object key) {
...
+        Duration duration;
         while ((obj = inp (key)) == null &&
-                (duration = Duration.between(now, Instant.now()).toMillis()) < timeout)
+                to.compareTo(duration = Duration.ofNanos(System.nanoTime()).minus(now)) > 0)
```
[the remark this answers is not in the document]
jameshilliard: You can't pass `System.nanoTime()` to `Instant` directly since `System.nanoTime()` is not a proper epoch but rather a runtime `Duration`. Basically an `Instant` needs to be passed `NanoClock` for monotonic behavior, but here since we only need to compute relative time we an just use a `Duration`.

jpos/src/main/java/org/jpos/q2/Q2.java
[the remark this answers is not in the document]
jameshilliard: From my understanding `System.nanoTime()` is the correct way to measure uptime since it's monotonic.

jpos/src/main/java/org/jpos/q2/cli/TZCHECK.java
```diff
@@ -35,7 +36,7 @@
     public void exec(CLIContext cli, String[] args) throws Exception
     {
         ZoneId zi = ZoneId.systemDefault();
-        Instant i = Instant.now();
+        Instant i = Instant.now(NanoClock.systemUTC());
```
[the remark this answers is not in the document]
jameshilliard: This is mostly for consistency.

jameshilliard:
> I pointed out some places where duration are used instead of Instants

That's intentional since `System.nanoTime()` is not an `Instant`(which is effectively an epoch based timestamp) but a `Duration`(effectively a runtime based timestamp), we only need to use `Instant` if we have API's that need epoch's.

jpos/src/main/java/org/jpos/space/TSpace.java
jameshilliard: I've refactored `TSpace` to not use `Instant` at all and only use monotonic `Duration`'s.

jpos/src/main/java/org/jpos/q2/qbean/SystemMonitor.java
```diff
@@ -185,7 +185,7 @@ private String generateFrozenDump(String indent) {
         String newIndent = indent + "  ";
         Runtime r = getRuntimeInstance();
         ZoneId zi = ZoneId.systemDefault();
-        Instant instant = Instant.now();
+        Instant instant = Instant.now(NanoClock.systemUTC());
```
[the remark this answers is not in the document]
jameshilliard: This is mostly for consistency.

jameshilliard:

PR #394, opened 2020-12-30
Add -XX:+IgnoreUnrecognizedVMOptions to bin/q2.

jameshilliard:


ar:
This is a nice JVM switch I wasn't aware it existed. It is good to know it's there, but a better approach is to be less opinionated about the options in jPOS regular startup scripts. That's a job for the ops team. I'll merge it now, but will take it away soon, along with other options that we don't necessarily need.

PR #402, opened 2021-03-09
Run tests on jdk16 and fix build+test warnings/errors.

jameshilliard:


jpos/src/main/java/org/jpos/security/jceadapter/JCESecurityModule.java
```diff
@@ -1823,7 +1823,11 @@ private void init (String jceProviderClassName, String lmkFile, boolean lmkRebui
                     evt.addMessage("No JCE Provider specified. Attempting to load default provider (SunJCE).");
                     jceProviderClassName = "com.sun.crypto.provider.SunJCE";
                 }
-                provider = (Provider)Class.forName(jceProviderClassName).newInstance();
+                if (jceProviderClassName == "com.sun.crypto.provider.SunJCE") {
```
rubieHess, line 1826: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [ES: Comparison of String objects using == or != (ES_COMPARING_STRINGS_WITH_EQ)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#es-comparison-of-string-objects-using-or-es-comparing-strings-with-eq).
This code compares java.lang.String objects for reference equality using the == or != operators. Unless both strings are either constants in a source file, or have been interned using the String.intern() method, the same string value may be represented by two different String objects. Consider using the equals(Object) method instead.

jameshilliard: Fixed, I seem to miss that quirk fairly often since most other languages have `==` operators that work differently.

Issue #414, opened 2021-05-27
MuxPool not honoring a timeout of 0


chhil:
MuxPool request method
https://github.com/jpos/jPOS/blob/master/jpos/src/main/java/org/jpos/q2/iso/MUXPool.java#L72-L82

If timeout is 0. (I don't want to set a mux timeout, let it take as much time as it wants).

```
long maxWait = System.currentTimeMillis() + timeout;
followed by 
timeout = maxWait - System.currentTimeMillis();
```

timeout will return a negative timeout (or on fast systems the  System.currentTimeMillis() may not change due to OS timer precision and result in timeout of 0).  Currently a null is being returned (see snippet below)

```
        if (mux != null) {
            timeout = maxWait - System.currentTimeMillis();// maxwait is the previous call to System.currentmilli
            if (timeout >= 0)
                return mux.request (m, timeout);
        }
        return null;
```

My thoughts on this are:
If a timeout passed to the muxpool.request is 0, means we don't care how much time is taken in the muxpool and the 0 timeout must be passed to the underlying mux. 

Additionally, when timeout of 0 is passed to the muxpool the call :
```MUX mux = getMUX(m,maxWait);```
The snippet below will result in the mux selection being called only once.
```
} while (System.currentTimeMillis() < maxWait); // maxwait is the system.currentmill and while loop will terminate after looping once as System.currentTimeMillis() will always be greater than maxWait.
```

So, a decision needs to be made on supporting a timeout of 0 and how long do we want to loop in getting the mux if none are available when timeout is 0.

chhil:
The above is based on a slack conversation
https://jpos.slack.com/archives/CHAB8RFKK/p1621959937026900

There is also muxpool.send issue stated by Jason Hilliard.
https://jpos.slack.com/archives/CHAB8RFKK/p1622066968035200?thread_ts=1621959937.026900&cid=CHAB8RFKK

It kind of states a similar thing to looping once to get a mux when the muxpool.send method is used.

jameshilliard:
Potential fix in #415.

PR #424, opened 2021-05-28
Use monotonic clocks for MUXPool and fix timeout calculation bug.

jameshilliard:
Alternative to #415.

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/iso/MUXPool.java
+++ b/jpos/src/main/java/org/jpos/q2/iso/MUXPool.java
@@ -27,6 +27,7 @@ import org.jpos.space.SpaceFactory;
 import org.jpos.util.NameRegistrar;
 
 import java.io.IOException;
+import java.time.Duration;
 import java.util.StringTokenizer;
 import java.util.concurrent.atomic.AtomicInteger;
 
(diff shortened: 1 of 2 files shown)


ar:
You mix stuff in the PRs which make them difficult to accept.

Can you reference which timeout calculation bug you are referring too?

jameshilliard:
> You mix stuff in the PRs which make them difficult to accept.

The other changes are somewhat related to the primary fix.

> Can you reference which timeout calculation bug you are referring too?

The main issue is with this code:
https://github.com/jpos/jPOS/blob/6c05df80b13d21d5686cddceca0db6d79aeceb09/jpos/src/main/java/org/jpos/q2/iso/MUXPool.java#L77-L79

Where the `if (timeout >= 0)` check is done against a recomputed `timeout` instead of the `timeout` originally passed to the function. The recomputed `timeout` additionally appears to result in inconsistent behavior when exactly 0 since `QMUX` requests appear to have special cased behavior for timeouts of zero here(and in a few other places I think):
https://github.com/jpos/jPOS/blob/6c05df80b13d21d5686cddceca0db6d79aeceb09/jpos/src/main/java/org/jpos/q2/iso/QMUX.java#L142-L145

ar:
I understand you are referring to issue #414 - we don't need nano precision and excessive object reaction to fix that.

In addition, your fix doesn't address the real intent of using a 0 timeout, which is fire-and-forget the message. In addition, this fix only fixes the synchronous call to MUX.request, not the async one.

I addressed this issue in 4cd6c957e - we can raise a case for the monotonic clock stuff separately.

jameshilliard:
> I understand you are referring to issue #414 - we don't need nano precision

Reason for using `System.nanoTime()` isn't precision related but rather due to `System.currentTimeMillis()` being unsafe for elapsed time calculations due to using the wrong type of clock. Note on Linux `System.nanoTime()` calls are actually faster than `System.currentTimeMillis()` calls.

> excessive object reaction to fix that.

There is some overhead from this but it seems to be minimal from some benchmarks I did, we could technically avoid it but not sure that would be worth it.

jameshilliard:
> In addition, your fix doesn't address the real intent of using a 0 timeout, which is fire-and-forget the message. In addition, this fix only fixes the synchronous call to MUX.request, not the async one.

Hmm, the async version still doesn't really look right to me. Reworked some of the logic here more in #425.

PR #425, opened 2021-05-29
Use monotonic clocks for MUXPool and fix send/timeout bugs.

jameshilliard:
This also avoids using Duration for performance. `System.nanoTime()` is faster than `System.currentTimeMillis()` on Linux so moving to monotonic clocks should also improve performance.


ar:
Thank you for the PR - see https://github.com/jpos/jPOS/commit/4cd6c957e443529f1a88c9af954f9c6432796b28#commitcomment-51470635

PR #441, opened 2021-08-09
Adding Q2 cmd line option to set the jpos.envdir

barspi:


jpos/src/main/java/org/jpos/core/Environment.java
```diff
@@ -69,6 +69,9 @@ private Environment() throws IOException {
     public String getName() {
         return name;
     }
+    public String getDir() {
```
ar, line 72: Given this is a public method, can we call it `getEnvDir()` instead of `getDir()` just to show we are talking about the 'environment' directory, and not Q2's CWD (which is actually the JVM's CWD), please?

jpos/src/main/java/org/jpos/q2/Q2.java
```diff
@@ -771,10 +785,7 @@ private void parseCmdLine (String[] args) {
                 pidFile = line.getOptionValue("p");
             if (line.hasOption("n"))
                 name = line.getOptionValue("n");
-            if (line.hasOption("E")) {
-                System.setProperty("jpos.env", line.getOptionValue("E"));
-                Environment.getEnvironment();
```
ar, line 776: So you removed the required initial call to `Environment.getEnvironment()` ?

jpos/src/main/java/org/jpos/core/Environment.java
barspi: no problem, easy change

I tought it was implicit by virtue of having to call it on the `Environment` instance (such as the pre-existing `getName()` and could save a few characters

jpos/src/main/java/org/jpos/q2/Q2.java
barspi: This call was being made only for option  `E`
I moved this code to the top, plus the one that parses the new option `Ed` (for the envdir)
Then I added
```java
if (line.hasOption("Ed") || line.hasOption("E")) {
    Environment.reload();
}
```

The `reload()` method is returning the `INSTANCE` as the `getEnvironment()` was doing, but only after reloading with the new settings.  I put that at the beginning, before the other cmd line options are parsed, to give it the opportunity to have the env loaded (and from the proper file!), before other parts of the system

If neither `E` nor `Ed` is explicited, whoever gets the environment first in the global code would be initializing the static `INSTANCE` anyway (from the default file).  That's still the same that was happening before.

I hope my logic is clear, and especially hope it's correct :-)

ar (approved):

ar (approved):

PR #448, opened 2021-10-06
generalize commaEncode/Decode in ISOUtil

pmarziotto-midinero:
generalize commaEncode/Decode in ISOUtil to use any character as delimeter


alcarraz:
great addition @pmarziotto-midinero, may I suggest renaming the methods to `separatorDecode/encode` or `delimiterDecode/Encode`? 

Or perhaps a better name I can't think of that may make more explicit the purpose of the method?

PR #455, opened 2021-11-23
add autoconfig support

ar:


chhil:
Curious why boolean is not supported.

What would happen if the type were an int but say the data is passed in as "XYZ"?
If I use the cfg.getInt(...) it would have thrown an error.
    ```
public int getInt (String name) {
        return Integer.parseInt(get(name, "0").trim());
    }
```

ar:
My bad - implemented now:
https://github.com/jpos/jPOS/commit/b5b96f97bbf7dc9c75c060bb7f08d1c7ac752b36

--
@apr <http://twitter.com/apr>

Issue #462, opened 2022-01-05
XXE vulnerability in GenericPackager


k-sever:
GenericPackager (as well as GenericValidatingPackager) are vulnerable to [XML External Entity (XEE) attack](https://owasp.org/www-community/vulnerabilities/XML_External_Entity_(XXE)_Processing).

Here is the test example that reproduces it 
```java
@Test
    public void testXXEAttach() {
        try {
            String xeeAttackXml = "<?xml version=\"1.0\" encoding=\"ISO-8859-1\"?>\n"
                + "<!DOCTYPE foo [\n"
                + "  <!ELEMENT foo ANY >\n"
                + "  <!ENTITY xxe SYSTEM \"https://jposxee.free.beeceptor.com/hacked\" >]>\n"
                + "<foo>&xxe;</foo>";
            new GenericPackager().readFile(new ByteArrayInputStream(xeeAttackXml.getBytes()));
            fail("Expected ISOException to be thrown");
        } catch (ISOException ex) {
            assertEquals("Entity extension is disabled", ex.getMessage(), "ex.getMessage()");
        }
    }
```
The `xxe` entity is getting resolved and request goes to `https://jposxee.free.beeceptor.com/hacked` (it's just a mock API):
<img width="1216" alt="Screenshot 2022-01-05 at 22 34 45" src="https://user-images.githubusercontent.com/8113027/148285444-1b422da0-a887-41d2-abc5-4c59c8b67e2c.png">

It's not a finalized test that you can check in, but it gives you the idea.

ar:
GenericPackager configuration is usually read from the classpath. Having the ability to define entities is a nice to have feature that enhances the code. Somebody with enough access to the classpath to use an XXE have easier ways to compromise the system. 

Can you provide a use case where a GenericPackager configuration could be injected from a remote host?

k-sever:
> GenericPackager configuration is usually read from the classpath.
Well, you can't enforce it. IMO it should be secure by default, security > code readability. I'd rather disable external entities and make it possible to configure XML reader features for GenericPackager, so that anyone who actually needs it can redefine it.

ar:
Fair enough. I'll reopen the issue and think about a way to configure it to reenable the feature.

PR #463, opened 2022-01-05
Fix XXE vulnerability

k-sever:
## Fixes https://github.com/jpos/jPOS/issues/462

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/packager/GenericPackager.java
+++ b/jpos/src/main/java/org/jpos/iso/packager/GenericPackager.java
@@ -251,6 +251,9 @@ public class GenericPackager
             );
         }
         reader.setFeature ("http://xml.org/sax/features/validation", true);
+        reader.setFeature("http://xml.org/sax/features/external-general-entities", false);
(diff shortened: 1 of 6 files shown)


jpos/src/main/java/org/jpos/iso/packager/GenericPackager.java
```diff
@@ -251,6 +251,9 @@ private XMLReader createXMLReader () throws SAXException {
             );
         }
         reader.setFeature ("http://xml.org/sax/features/validation", true);
+        reader.setFeature("http://xml.org/sax/features/external-general-entities", false);
+        reader.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+        reader.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
```
k-sever, line 256: https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html

Agitolyev:
Hi, is there a plan to merge this PR?

ar:
This PR needs a way to revert, via configuration, to the old behavior, just in case there's an use-case for entity expansion.

Issue #264, opened 2019-12-10
RotateLogListener timezone


ar:
Galih reports:

```
hi guys,
it seems like the log time at time exceeded - log rotated and iso msg seems like have a different format time
<?xml version="1.0" encoding="UTF-8"?>
<logger class="org.jpos.util.DailyLogListener">
<log realm="rotate-log-listener" at="Mon Dec 09 00:00:14 WIB 2019">
  maxage feature is disabled.
</log>
<log realm="rotate-log-listener" at="Tue Dec 10 00:00:21 WIB 2019">
  time exceeded - log rotated
</log>
</logger>
~
<log realm="channel/10.100.111.181:7990" at="2019-12-10T09:45:16.937">
 <send>
   <isomsg direction="outgoing">
     <!-- org.jpos.iso.packager.GenericPackager[iso_spec_pce.xml] -->
     <field id="0" value="0800"/>
     <field id="7" value="1210094516"/>
     <field id="11" value="088717"/>
     <field id="70" value="001"/>
   </isomsg>
 </send>
</log>
```

Alcarraz points out we use this (new Date()) at rotate time:

```
 protected synchronized void logDebug (String msg) {
        if (p != null) {
            p.println ("<log realm=\"rotate-log-listener\" at=\""+new Date().toString() +"\">");
            p.println ("   "+msg);
            p.println ("</log>");
        }
    }

but

```
sb.append("\" at=\"");
sb.append(LocalDateTime.ofInstant(dumpedAt, ZoneId.systemDefault()));
```

on regular events.

We need to use ZoneId at rotate time too.

PR #478, opened 2022-04-16
Update RotateLogListener.java

galihlasahido:
update for issue https://github.com/jpos/jPOS/issues/264 regarding need to use ZoneId at rotate time because it's different on regular events.


ar:
closes #264

PR #496, opened 2022-08-17
feat: Support for 8 length bines on ISOUtil

### LINKED ISSUE #475: Masking support for 8-digit BINs
[ISOUtil#protect()](https://github.com/jpos/jPOS/blob/c5735364e910e8b552924038660bc2c11f2ec2b6/jpos/src/main/java/org/jpos/iso/ISOUtil.java#L903) should have configurable/dynamic masking of PANs to support 8-digit BINs

Come up with a flexible and backward-compatible (default) solution.

aVolpe:
Configured with the environment property ${jpos.util.bin.length}, it
should be a comma separated list of bines.

For example, if there are two 8 length bines "4000000" and "2000000",
then this property should be:

 jpos.util.bin.length = 4000000,2000000

fixes #475

Signed-off-by: Arturo Volpe <avolpe@fintech.works>

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/ISOUtil.java
+++ b/jpos/src/main/java/org/jpos/iso/ISOUtil.java
@@ -18,6 +18,7 @@
 
 package org.jpos.iso;
 
+import org.jpos.core.Environment;
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.Charset;
@@ -26,12 +27,17 @@ import java.text.DecimalFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.BitSet;
+import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Random;
+import java.util.Set;
 import java.util.StringJoiner;
 import java.util.StringTokenizer;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+
 
 /**
  * various functions needed to pack/unpack ISO-8583 fields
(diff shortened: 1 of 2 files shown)


ar:
According PCI:

The masking approach should always ensure that only the minimum number of digits is displayed as necessary to perform a specific business function. For example, if only the last four digits are needed to carry out a business function, mask the PAN so that individuals performing that function can view only the last four digits. While the intent of Requirement 3.3 is to display no more than the **“first six and last four digits”** of a PAN, an entity will be permitted to display more digits if needed **but only with a documented business justification**.

The solution you provide Arturo is good, backward compatible, and easy to rollout, but we need to be aware that a documented business justification to configure these extended BINs. People may really don't need it.

aVolpe:
After a bit of googling, I found this page: https://src-gmbh.de/en/8-digit-bins-and-pci-dss/ 

It says that in PCI DSS v4.0 that requirement changes to:

> The current summary in the PCI SSC FAQ entry now defines that “first 8, any other 4” is permitted for truncation for 16-digit PANs. The (test) card number 4012888888881881 is then allowed to be stored and processed in the form 40128888xxxx1881, for example — it is sufficient if any four digits are crossed out after the BIN. Only for shorter PANs, the existing rules “first 6, any other 4” (Discover) or “first 6, last 4” (American Express) remain in place. A corresponding adjustment of the PCI DSS requirement for masking is expected with the [change to PCI DSS v4.0](https://src-gmbh.de/en/pci-dss-v4-0-is-coming-an-overview/).

So we can use this solution if we are aiming at PCI DSS v4.0, but I don't have the actual document so I can't confirm the validity of that information.

aVolpe:
The spec https://docs-prv.pcisecuritystandards.org/PCI%20DSS/Standard/PCI-DSS-v4_0.pdf in section 3.4.1

> 3.4.1 PAN is masked when displayed (the BIN and
last four digits are the maximum number of digits
to be displayed), such that only personnel with a
legitimate business need can see more than the
BIN and last four digits of the PAN.

But it doesn't say the length of the bin. And I can't find any reference to the length of the bin.

aVolpe:
This FAQ further clarifies the masking: https://www.pcisecuritystandards.org/faq/articles/Frequently_Asked_Question/What-are-acceptable-formats-for-truncation-of-primary-account-numbers/

This PR is compatible with that change.

The only thing that I don't like is the name of the property, but I can't think of a better name

PR #512, opened 2023-02-22
rc.CMF: add CONFIGURATION_ERROR, plus fixing typos and enum conflicts

barspi:
* Adding new CMF constant  CONFIGURATION_ERROR (as per jPOS-CMF.pdf)
* Fixing typos in some constant names
* By fixing the spelling in `SUSPECTED_COUNTERFELT_CARD(2010)`, a conflict was noticed because a correctly-spelled constant was already defined under code `1029`.  After consulting with v2003 it was determined that in one case "card pickup" was required and not in the other, so the constants were renamed accordingly.


jpos/src/main/java/org/jpos/rc/CMF.java
```diff
@@ -67,7 +67,7 @@ public enum CMF implements IRC {
     INVALID_PINBLOCK(1026),
     PIN_LENGTH_ERROR(1027),
     PIN_KEY_SYNC_ERROR(1028),
-    SUSPECTED_COUNTERFEIT_CARD(1029),
+    SUSPECTED_COUNTERFEIT_CARD_NOT_PICKUP(1029),
```
aVolpe, line 70:
```suggestion
    SUSPECTED_COUNTERFEIT_CARD_DONT_PICKUP(1029),
```

The intention of the message is that there is a suspected fraud, but we shouldn't seize the card.

jpos/src/main/java/org/jpos/iso/ISOMsg.java
```diff
@@ -285,32 +285,34 @@ public void set(String fpath, String value) {
...
+                     */
+                    if (c == null) {
+                        break;
+                    } else {
+                        // We have a value to set, so adding a level to hold it is sensible.
+                        m.set(m = new ISOMsg(fldno));
+                    }
+            } else {
+                if (c != null)
+                    c.setFieldNumber(fldno);
```
alcarraz, line 310: If  `c` is null, wouldn't be better to not add the inner fields in the first place, and just return at the beginning of the method?
The current modification is better for maintaining secondary effects for backward compatibility, but it looks weird to me.

barspi: @alcarraz something weird happened
The change in `ISOMsg` should have been a completely separate PR from the one fixing `rc.CMF`

You are more or less right.  The semantics for `set(String, ISOComponent)` are not exactly the same as the semantics for other versions of `set()`.
For example, for a `m.set("95.2.8", (iSOComponent) null)  I would expect to be equivalent to `m.unset("95.2.8")` 

This change was specifically to fix the case of, for example, an inner `ISOAmount` that had no internal id (`fieldNumber`, or `key`) **set beforehand**, or it had one but needed to be explicited again in the fpath, and be the **same number**, otherwise the hierarchy would not be correctly created.

I tried not to break (too much!) the non-pathological usages of the method.
I'm sure that any code in existence has gone through the trouble of doing

```java
am = new ISOAmount(8);  // make the component's id explicit
m.set("95.2.8", am);    // express the whole path with 8 at the end
```

To me the semantics of `set(fpath, xxx)` should be "set xxx as the value of the last part of the path", so I shouldn't need to make the id explicit in the first line (could be -1, uninitialized, or maybe another value because the component was extracted from somewhere). Once it's inserted into the correct point, it's id should be **updated** (actually, **before** it's inserted, because that's how `m.set(c)` works.

barspi:
I just reverted the unrelated `ISOMsg` commits and will send a separate PR

jpos/src/main/java/org/jpos/rc/CMF.java
barspi: @aVolpe your suggestion has been applied

jpos/src/main/java/org/jpos/iso/ISOMsg.java
alcarraz: Sorry, I have misread the code change, I missed that you maintained the next `m.set` so I believed your intention was what I said previously, my comments have no sense whatsoever with the real change, and the comment would be another, that I will make in the new PR.

barspi: There’s a new PR for that already, with better explanation and tests, if you want to check it out 😁

ar (approved):

PR #520, opened 2023-04-18
Basic support for string array auto-configuration.

alcarraz:
Proposal for a basic support for autoconfiguration of string array fields. 

More complex logic could be added to support different ways of defining string arrays, or even collections of other types, in a single property, if desired, but this could do for now.

An unrelated fix for the doc is also included, that may not deserve a specific PR, and updating my mail address also.

If preferred, I can break this PR in 2 or 3.

### DIFF
--- a/CREDITS
+++ b/CREDITS
@@ -11,7 +11,7 @@ Hani S. Kirollos <h.samuel-at-egyptianbanks.net>
 Jonathan O'Connor <Jonathan.O'Connor-at-xcom.de>
 Cameron Young <cn_young-at-attcanada.ca>
 Alireza Taherkordi <taherkordy-at-dpi2.dpi.net.ir>
-Alcarraz, Andrés <alcarraz-at-jpos.org>
+Alcarraz, Andrés <alcarraz-at-gmail.com>
 Alwyn Schoeman <alwyn.schoeman-at-gmail.com>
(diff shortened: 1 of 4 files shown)


ar (approved):

ar:
Good PR I was about to implement. It would be great to have support for `int[]` and `long[]` as well. Will add it as time permits.

alcarraz:
Yes, sorry.

For a property definition like this:
```xml
<property name="array" value="one two three"/>
```

Option 1 would be something like this:
```java
@Config(value="array", multi=false, separator=' ') 
```

Of course, the `multi` field could be single and have the opposite meaning, depending on what feels clearer.

Option 2:
```java
@Config(value="array", separator=" ")
```

It could also be something like this:

```java
@Config(value="array", separators=", ") //in this case the name should be plural to better indicate intention
```

It also has to be decided which should be the sane defaults.

ar:
I don't think we want to create a DSL here. What I suggest for the time being is to support the equivalent of `cfg.getInts()` and `cfg.getLongs()`.

alcarraz:
I was thinking about supporting the common case of space and comma separated arrays and collections, like in `CheckFields`. It's to avoid the verbosity in the configuration of having a `property` tag for each item.

PR #554, opened 2023-07-20
Enable environment property replacement anywhere

alcarraz:
This draft PR is a possible implementation of what's discussed in #546.

When an element is instantiated via `QFactory.instantiate` the element is passed to the added `replaceEnvProperties` method, which replaces all environment properties recursively on its children and its attributes.

If an element has the `verbatim` attribute set to `true` it will not process it or its children, I'm inclined to think this should be the correct behavior because it also acts as a performance hint to not recursively process its children. However, one may expect to override the parent's option in a child, so this is debatable. This also simplifies the code, by not having to track changes in the attribute down the line in the recursion. 

It may be necessary to remark that the `verbatim` option does not apply to the `Configuration` elements or other classes that call `Environment.getProperty` on its own. This may sound obvious, but some may get confused about that.

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/Q2.java
+++ b/jpos/src/main/java/org/jpos/q2/Q2.java
@@ -49,7 +49,6 @@ import javax.crypto.spec.SecretKeySpec;
 import javax.management.InstanceAlreadyExistsException;
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanServer;
-import javax.management.MBeanServerFactory;
 import javax.management.ObjectInstance;
 import javax.management.ObjectName;
 import java.io.*;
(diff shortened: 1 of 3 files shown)


alcarraz:
I removed the draft flag, because I added the verbatim tests.

However, I'm not entirely convinced on performing the replacement in the `instantiate` method instead of in `setConfiguration`. This let us avoid multiple passes when calling `setConfiguration`  on children, but would not catch eventual `setConfiguration` calls with XML fragments taken from other places.

alcarraz:
Once this PR is merged, I commit myself to add something on the programmer's guide, documenting this.

ar (approved):

PR #559, opened 2023-08-09
Allow setting of a higher `corePoolSize` while creating new `ScheduledThreadPoolExecutor`

sumeetphadnis:
At present `ConcurrentUtil.newScheduledThreadPoolExecutor()`, which is called from `QBeanSupport.getScheduledThreadPoolExecutor()`, hard-codes `corePoolSize` to 1. This change allows us to set a higher value via a configuration property named `stpe-core-pool-size`

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/QBeanSupport.java
+++ b/jpos/src/main/java/org/jpos/q2/QBeanSupport.java
@@ -255,8 +255,9 @@ public class QBeanSupport
     protected void destroyService() throws Exception {}
 
     protected synchronized ScheduledThreadPoolExecutor getScheduledThreadPoolExecutor() {
+        int corePoolSize = cfg.getInt("stpe-core-pool-size", 1);
         if (scheduledThreadPoolExecutor == null)
-            scheduledThreadPoolExecutor = ConcurrentUtil.newScheduledThreadPoolExecutor();
+            scheduledThreadPoolExecutor = ConcurrentUtil.newScheduledThreadPoolExecutor(corePoolSize);
         return scheduledThreadPoolExecutor;
     }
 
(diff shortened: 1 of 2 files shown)


jpos/src/main/java/org/jpos/q2/QBeanSupport.java
```diff
@@ -255,8 +255,9 @@ protected void stopService()    throws Exception {}
     protected void destroyService() throws Exception {}
 
     protected synchronized ScheduledThreadPoolExecutor getScheduledThreadPoolExecutor() {
+        int corePoolSize = cfg.getInt("stpe-core-pool-size", 1);
```
alcarraz, line 258: What if we just add the number of threads to be a parameter passed by the caller?

We could still use this default (mixinig feelings about if we should) in the no args version, but this would allow for more control in the participant's code.

alcarraz: Sorry, my bad, this is initializing a member variable, so it's not a good idea. Participant can already override the method.

sumeetphadnis: We can also have an overridden method in QBeanSupport to accept a pool size argument. If that is a better way, I can change it accordingly. This approach of reading a cfg property will also work, but one draw back is that the name of the property is hard-coded.

alcarraz: Another option would be to delegate the initialization or the call to a getter of the pool size to a hook method. The implementation could read the property value, and the participant can choose to get the value from what ever it wants, it could even be a function of other properties or system config.

I believe this could be the sweet spot of configurability vs code complexity.

ar (approved):

Issue #568, opened 2023-11-16
using BERTLVAsciiHexPackager and AsciiHexInterpreter returns incorrect  uninterpretLength 


Elias-Terranti:
when using BERTLVAsciiHexPackager  to unpack a tlv , it unpacks the tag correctly but  when the packager calls on getUninterpretLength(len, AsciiHexInterpreter.instance ), the return is always 0, so the value is  an empty array

` my test code:

    public class jposbugtest {

    final String HEX_ASCII = "9f2610e18d2a69a45dd09a9f360401839f3708ab34b0bd";

    @Test
    public void testLenInterpretation() throws ISOException {
        final ISOMsg ICC_DATA = new ISOMsg(55);
        final BERTLVAsciiHexPackager packager = new BERTLVAsciiHexPackager();
        final IFA_LLABINARY fieldPackager = new IFA_LLABINARY();

        packager.setFieldPackager(new org.jpos.iso.ISOFieldPackager[]{fieldPackager});
        ICC_DATA.setPackager(packager);

        ICC_DATA.unpack(HEX_ASCII.getBytes());

        ICC_DATA.getChildren().values().forEach
                (e -> {
                    try {
                        ISOTaggedField field = (ISOTaggedField) e;
                        final byte[] value = (byte[]) field.getValue();
                        System.out.println("tag "+ field.getTag() + " value " + ISOUtil.hexString(value));

                        Assert.assertFalse(Arrays.isNullOrEmpty(value));

                    } catch (ISOException ex) {
                        throw new RuntimeException(ex);
                    }
                });

    }

}`

i tested using  both java 8 and java 17 with same issue

ar:
This TLV payload doesn't seem correct `9f2610e18d2a69a45dd09a9f360401839f3708ab34b0bd`. Tag 9f26 is OK, but tag `ab` is missing data.

```xml
    <tlvlist>
       <tag id='9f26' value='E18D2A69A45DD09A9F360401839F3708' />
    </tlvlist>
```

ar:
The test is not correct. Try this code:

```java
    @Test
    @SuppressWarnings("unchecked")
    public void testLenInterpretation() throws ISOException {
        final String HEX = "9f2610e18d2a69a45dd09a9f360401839f3708"; //     ab34b0bd";

        Logger logger = new Logger();
        logger.addListener(new SimpleLogListener(System.out));

        final ISOMsg ICC_DATA = new ISOMsg(55);
[...]

PR #570, opened 2023-11-23
issue #568: fixed BERTLVAsciiHexPackager  getUninterpretLength always returns 0 with AsciiHex interpreter

Elias-Terranti:
The issue was due to The two operands  lengthAdjusted  and interpreter.getPackedLength(lengthAdjusted) are both Integers.

the result was calculated using Integer arithmetic resulting in a 0,5 that was always rounded down to 0.  

0.5 because AsciiHex packed len is double len 

solution is to cast one of the operands to a double to use double arithmetic, then cast result to integer.

### DIFF
--- a/jpos/src/main/java/org/jpos/tlv/packager/bertlv/BERTLVPackager.java
+++ b/jpos/src/main/java/org/jpos/tlv/packager/bertlv/BERTLVPackager.java
@@ -447,7 +447,7 @@ public abstract class BERTLVPackager extends GenericPackager {
     private int getUninterpretLength(int length, BinaryInterpreter interpreter) {
         if (length > 0) {
             int lengthAdjusted = length + length % 2;
-            return (int) (length * (lengthAdjusted / (double) interpreter.getPackedLength(lengthAdjusted)));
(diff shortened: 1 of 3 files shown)


ar:
Thank you for the PR and detailed Unit test. 

There's no need for the double conversion, @barspi suggest a simpler solution, just

`return (length * lengthAdjusted) / interpreter.getPackedLength(lengthAdjusted);`

While you are at it, it would be great to also change the following method:

```
 private int getUninterpretLength(int length, Interpreter interpreter) {
        if (length > 0) {
            int lengthAdjusted = length + length % 2;
            return (length * lengthAdjusted) / interpreter.getPackedLength(lengthAdjusted);
        }
        return 0;
    }
```

(the one that uses an `Interpreter` instead of a `BinaryInterpreter`.

Would you like to send a new PR our of your branch or you prefer us to do it? Please confirm.

Elias-Terranti:
Hello @ar 

I've pushed a new commit with the  suggested edits , you should be able to see it added to the PR.

ar (approved):

PR #572, opened 2023-11-27
Add protected method to connect channel to streams

alcarraz:
These methods can be overridden by subclasses of existing channels to write to/read from streams,
for example, to replay some captured stream,
or just to do some tests based on ByteArrayInputStream.

Sometimes, we may want to use the existing channels,
but not necessarily to connect them to a TCP/IP endpoint,
for instance, we may want to input data from a file directly as an input of the channel.

Here, there is an example I'm using to read XML messages from a file or standard input:

```java
public class XMLInputFileChannel extends XMLChannel implements AutoCloseable{
    public XMLInputFileChannel() throws ISOException {
        this(System.in);
    }
    public XMLInputFileChannel(InputStream in) throws ISOException {
        this(in, new XMLPackager());
    }

    public XMLInputFileChannel(XMLPackager packager){
        this(System.in, packager);
    }

    public XMLInputFileChannel(InputStream in, XMLPackager packager) {
        connect(in, null);
        setPackager(packager);
        
    }

    @Override
    public void close() throws IOException {
        disconnect();
    }

    @Override
    public boolean isConnected() {
        return super.isConnected() || (usable && serverIn != null);
    }
}
```

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/BaseChannel.java
+++ b/jpos/src/main/java/org/jpos/iso/BaseChannel.java
@@ -291,6 +291,20 @@ public abstract class BaseChannel extends Observable
         setChanged();
         notifyObservers();
     }
+
+    /**
+     * Allow the channel to be connected to a pair of input and output streams.
+     * <p> 
+     * For instance for taking input from or outputting to a file or standard input/ouput.  
+     * 
+     * @param in Where to read from.
+     * @param out Where to write to.
+     */
+    protected void connect(InputStream in, OutputStream out) {
+        usable = in != null || out != null; //at least one of them has to be not null
+        if (in != null) serverIn = new DataInputStream(in);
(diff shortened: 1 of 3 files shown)


ar:
It would be nice to have this `XMLInputFileChannel` as part of the PR and it wouldn't hurt to have at least a unit test so that we can make sure we don't break it in the future with changes to BaseChannel.

alcarraz:
Sorry, I had written a couple of tests, but somehow the test class got stashed and I didn't notice.

I don't think it is a good idea to add the `XMLInputFileChannel`, because it is for quite some narrow use case (where connecting in the constructor was justified). However, I understand it wouls be good to have an example usage. So, I'll add a more generic on a followup commit to this PR as soon as I can write it. 

Thank you.

PR #577, opened 2023-12-21
Added a padding option to cryptogram builder

T-eli:
**changes:**

- added Padding Method interface 

-  added a buildARQCRequest method with PaddingMethod paramater  to return Cryptograms data with  specified padding 

-  added CPA or Common Payment Application Cryptogram Specs 

-  added fix for IssuerApplicationData does not match if iad is lowercase

- added tests

- fix typos/text 

**note:** 

i am not entirely sure about the getDefaultARPCRequest return for CPA CryptogramSpecs so i set it to Ascii Hex String 3030 for approved and 3031  for declined.


T-eli:
I have also added  ISO9797 Padding Method 3 in the Latest commit

jpos/src/main/java/org/jpos/emv/IssuerApplicationData.java
```diff
@@ -70,7 +70,7 @@ public IssuerApplicationData(String hexIAD) {
             //Therefore, the length of the Issuer Application Data is 18 bytes and for 
             // M/Chip Advance it may be 18, 20, 26, or 28.
             unpackMCHIP(iad);
-        } else if (len == 32 && iad.startsWith("0F") && iad.startsWith("0F", 32)) {
+        } else if (len == 32 && (iad.startsWith("0F") || iad.startsWith("0f")) && (iad.startsWith("0F", 32) || iad.startsWith("0f", 32))) {
```
barspi, line 73: A few stylistic suggestions:
* you may want to do a `toUpperCase()` (or lower...) after the `trim()` in line 60, to simplify the expression here.
* Renaming the variable `len` to `byteLen` will clarify you're not talking about the string but the bytes.
* writing `16*2` instead of `32` will make it clear as well (it confused us for a minute), and the compiler is smart enough to make it a constant 😉

T-eli: >you may want to do a toUpperCase() (or lower...) after the trim() in line 60 

this was my initial thought, but it ended up breaking some tests so  this is minimal way of fixing it.

- I was thinking of making the padding method  to take the Block size  n  as an argument as well, so it would be n*2  instead and it should work with 16 byte AES blocks as well. but I need to test it first.

PR #586, opened 2024-03-11
fix: ChannelAdaptor disconnect lock now a new Object. Fixes #585

### LINKED ISSUE #585: ChannelAdaptor potentially unsafe synchronization
Hello everyone,

Skimming through ChannelAdaptor source it seems in this [line](https://github.com/jpos/jPOS/blob/da7ccae557ce0931f73829b32fffdc89856461ee/jpos/src/main/java/org/jpos/q2/iso/ChannelAdaptor.java#L63) the disconnectLock is the shared Boolean.TRUE object.

This might be unsafe as someone might inadvertly deadlock the shutdown of a ChannelAdaptor by locking on Boolean.True. Also this would make the lock static in scope given that this lock instance would be shared among all other ChannelAdaptors.

Finally synchronizing against Boolean is warned against by the [Java spec](https://docs.oracle.com/javase/8/docs/api/java/lang/doc-files/ValueBased.html) as Boolean and a few other java.lang classes are considered ValueBased classes. (See @jdk.internal.ValueBased)

> A program may produce unpredictable results if it attempts to distinguish two references to equal values of a value-based class, whether directly via reference equality or indirectly via an appeal to **synchronization**, identity hashing, serialization, or any other identity-sensitive mechanism

Seems it could be swapped for `new Object()` without any side effect as far as I could see.

Probably never harmed anyone but better be safe!

Cheers

agustiza:


barspi:
Doig so e git archeology, it used to be a generic private Object but was changed into TRUE
https://github.com/jpos/jPOS/commit/301e36a603203df8113e5eba01de97607cd456b7

The explanation in the commits mentions that the Object would complicate using a persistent space such as JESpace (not much explanation as to how.. serialization of a ChannelAdaptor? and the comment itself shows some doubts about the usefulness)

I agree that the boolean is troublesome… maybe a string object would be the best of both worlds?

agustiza:
Good find!

Still I'm not sure the intention was to share the lock over a space (and it wouldn't really make sense if serialized over the wire). 

Seems like the intention was to simply use a boolean for the space signaling and the lock was included under a overly broad replace.

alcarraz:
It seems that the `disconnectLock` variable, is only used in a `synchronized` block, wouldn't be better to fix this with a reentrant lock?

agustiza:
Sure, they are pretty much equivalent in this case with the added benefit of being Loom proof should the sender or receiver become green threads in the future.

```
    protected void disconnect () {
        // do not synchronize on this as both Sender and Receiver can deadlock against a thread calling stop()
        disconnectLock.lock();
        try {
            SpaceUtil.wipe(sp, ready);
            channel.disconnect();
        } catch (Exception e) {
            getLog().warn("disconnect", e);
        } finally {
            disconnectLock.unlock();
        }
    }
```

PR #626, opened 2024-11-22
Add reset method to Metrics

mchhil-incomm:
Continuous counts at times may need a reset for grabbing values per iteration.
e.g. pull metrics info every minute, consume them and reset it , so that the metrics hold info for the poll interval.


mchhil-incomm:
Please ignore this one.

Use the next one

mchhil-incomm:
Please use mchhil-incomm:patch-3 PR

PR #639, opened 2025-05-08
1. bump versions of `micrometercore` and `micrometerprometheus` and fixing breaking changes

jxfzzzt:
1. bump versions of `micrometercore` and `micrometerprometheus` from 1.12.5 to 1.13.13.
2. fixing breaking changes and make sure all test cases pass.


jxfzzzt:
@ar @op-transactility  please review this dependency upgrade

ar:
Thank you for the contribution. I suggest to move to 1.14.6. Any reason not to do that?

ar:
Valid point @jxfzzzt but this is the `tip` version, bleeding edge, we try to go with the latest & greatest. Latest is always better. I went ahead and moved to 1.14.6 in commit 90ddadfe3 based on your PR. You saved me time with the package name change that I was not aware of. Thank you!

PR #641, opened 2025-07-18
Fix FSDMsg pack on field with separator to not unpad

mchhil-incomm:
If the new attribute pack_unpad with value false is added to the field, pack will not unpad (remove trailing spaces from the field).

If the attribute is absent or true backward compatibility is maintained by unpadding the field i.e. stripping the trailing spaces.

### DIFF
--- a/jpos/src/main/java/org/jpos/util/FSDMsg.java
+++ b/jpos/src/main/java/org/jpos/util/FSDMsg.java
@@ -247,6 +247,11 @@ public class FSDMsg implements Loggeable, Cloneable {
         return pack().getBytes(charset);
     }
 
+	    protected String get (String id, String type, int length, String defValue, String separator)
+        throws ISOException
+    {
+	    return get(id,type,length,defValue,separator,true);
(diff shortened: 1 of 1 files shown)


mchhil-incomm:
[FSDMsg_Field_Separator_Pack_TestCase_Backward_Compatible-base.xml.txt](https://github.com/user-attachments/files/21310313/FSDMsg_Field_Separator_Pack_TestCase_Backward_Compatible-base.xml.txt)
[FSDMsg_Field_Separator_Pack_TestCase_Fixed_Padding-base.xml.txt](https://github.com/user-attachments/files/21310314/FSDMsg_Field_Separator_Pack_TestCase_Fixed_Padding-base.xml.txt)
[FSDMsgFieldSeparatoPackTestCase.java.txt](https://github.com/user-attachments/files/21310315/FSDMsgFieldSeparatoPackTestCase.java.txt)


These are my test case files used for testing the change.

ar (approved):

PR #646, opened 2025-09-22
Correct use of serverOutLock in some channels

barspi:
In jPOS 3, `BaseChannel.serverOutLock` is now a `ReentrantLock`.
There was some older code trying to use it as a synchronized monitor object.

This PR fixes the usage for `ASCIIChannel` and a some minor fixes for `CSChannel` and  `AmexChannel` (includes a lot of indentation changes form tabs to space, sorry... but you can ignore whitespace changes)

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/channel/ASCIIChannel.java
+++ b/jpos/src/main/java/org/jpos/iso/channel/ASCIIChannel.java
@@ -113,9 +113,12 @@ public class ASCIIChannel extends BaseChannel {
                 if ((l=Integer.parseInt(new String(b))) == 0 &&
                     isExpectKeepAlive()) {
 
-                    synchronized (serverOutLock) {
+                    serverOutLock.lock();
+                    try {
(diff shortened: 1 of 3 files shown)


alcarraz:
This may not be the MR to perform this, but we have the same pattern in almost all channels, what if we add a `writeAndFlush(byte  [])` method in `BaseChannel` that performs the lock and does the writing and flushing?

This would improve consistent usage in all channels and lower the chances of misusage.

alcarraz:
> Maybe your `writeAndFlush(byte[])` could be used by `send(byte[])` . In fact, I just noticed a very specific one `sendKeepAlive()`, but not all channels have historically implemented the API the same way because of... well... historical reasons 😉

Yeah, my idea was for it to be called where this pattern appears. We don't have to replace every usage yet, it can be added in this MR and only used by these channels, and later modify in other places. But I think I would replace in every place the pattern appears unmodified.

ar (approved):

PR #701, opened 2026-04-12
fix(IFB_BITMAP): pack/unpack asymmetry when bit 1 is set by a sub-field data field

lyang:
## Problem

`IFB_BITMAP(length=16)` has a pack/unpack asymmetry when bit 1 of the inner BitSet is set by a real sub-field data field rather than by an explicit ISO-8583 secondary-bitmap extension indicator.

**Root cause:**
- `pack()` uses the formula `(b.length()+62 >>6 <<3)` — for sub-fields 1–6 present this gives **8 bytes**
- `unpack()` calls `ISOUtil.byte2BitSet(b, offset, 128)` which checks whether the MSB of `byte[0]` is set; since sub-field 1 being present naturally sets that bit, unpack **consumes 16 bytes** — 8 more than pack produced

The resulting 8-byte offset misalignment causes subsequent fixed-length field reads to run past the end of the sub-message buffer, typically as:
```
java.lang.StringIndexOutOfBoundsException: Range [45, 45 + 40) out of bounds for length 78
```

**Affected packager / field:**
- `cmf.xml` and `cmf-858.xml` — **DE-049** (Verification Data): `IFB_BITMAP(16)` with sub-field `id=1` ("Additional Identification Type") as a real data field
- Only `length >= 16` is affected; `length=8` is immune because `Math.min(8, 16) = 8` clamps consumed to 8 regardless of bit 1

Note: DE-043 (Card Acceptor Name/Location) is **not** affected because its data fields start at `id=2`, so bit 1 is never set by a data field.

## Failing test

Added `testLength16PackUnpackAsymmetryWhenBit1SetByDataField` to `IFB_BITMAPTest` — currently fails with:
```
expected: <8> but was: <16>
IFB_BITMAP(16) unpack consumed 16 bytes but pack only wrote 8; pack/unpack are
asymmetric when bit 1 is set by a data sub-field rather than an explicit extension indicator
```

This PR adds only the failing test to document the bug. A fix is not included — the correct remedy (extend pack to write 16 bytes when bit 1 is set, or reserve bit 1 as a non-data extension indicator in the XML) should be decided by the maintainers.

### DIFF
--- a/jpos/src/test/java/org/jpos/iso/IFB_BITMAPTest.java
+++ b/jpos/src/test/java/org/jpos/iso/IFB_BITMAPTest.java
@@ -25,6 +25,7 @@ import java.util.BitSet;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.DisplayName;
 
 /**
  * @author joconnor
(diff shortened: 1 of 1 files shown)


ar-agt:
I reproduced the failure locally, and the diagnosis matches the current `IFB_BITMAP.pack()` / `unpack()` behavior.

That said, this PR adds only a failing test and no fix, so I would not merge it as-is.

The key design question is whether the fix belongs in `IFB_BITMAP` itself, or in the packager definitions (`cmf.xml` / `cmf-858.xml`) where subfield `id=1` is being used as a real data field even though bit 1 is also the bitmap extension indicator.

I’d recommend either:

1. extending this PR with the actual fix, or
2. retitling it as a reproducer / draft and opening an issue to decide the correct behavior first.

ar-agt:
Opened #703 with the actual fix.

The root cause is in the CMF packager definitions: DE49 was configured with `IFB_BITMAP(16)` in `cmf.xml` and `cmf-858.xml`, but it should be 8 bytes because subfield 1 is real data and must not collide with the secondary-bitmap extension bit.

`cmfv3.xml` uses `DatasetPackager`, so the new PR only adds a clarifying comment there.

I think we can close this reproducer PR in favor of #703.

ar:
Thank you for the PR.

The failing test is valid, but the practical issue is in the CMF XML, not in `IFB_BITMAP` generally.

In DE-049, the inner bitmap is defined as `IFB_BITMAP(16)`, but DE-049 only has subfields 1..6, so it only needs an 8-byte bitmap. The problem is that subfield `1` is real data (“Additional identification type”).

With `IFB_BITMAP(16)`, `pack()` writes 8 bytes when only low bits are set, but `unpack()` interprets bit 1 as “secondary bitmap present” and consumes 16 bytes.

That creates a pack/unpack asymmetry and shifts the remaining field offsets.

  - `IFB_BITMAP(16)`: pack = 8, unpack = 16 in those situations
  - `IFB_BITMAP(8)`: pack = 8, unpack = 8

The right fix is to change DE-049’s bitmap from length 16 to length 8 in `cmf.xml` and `cmf-858.xml`, rather than changing `IFB_BITMAP` behavior globally.

The fix has been implemented in PR #703 so I'm closing this one. Feel free to reopen or file an issue if you think this needs further review.

PR #709, opened 2026-04-28
ProtectDebugInfo: enhanced config

barspi:
The original `ProtectDebugInfo` can be configured with **individual, repeatable properties** for all the elements we want to protect or wipe (context entries, iso fields, etc.) as shown in the javadoc.

This is nice but a bit verbose, and in particular does not allow to easily use build-time replacement configs, or run-time configurations (such as given in a jPOS env variable).

This PR adds the possibility for a protect/wipe property to also accept a comma/space-separated list of values.
This does not replace the current multi-property configuration but it's **in addition to**.
So backwards compatibility is maintained while allowing for other options.

For example, we can say:
```xml
<property name="protect-entry" value="RESPONSE" />
<property name="protect-entry" value="PAN, EXP, REQUEST_ICC_DATA" />
```

And both sets would be merged into one, giving the equivalent of having configured every `protect-entry` separately  (the classic way), or all in one line such as
```xml
<property name="protect-entry" value="RESPONSE, PAN, EXP, REQUEST_ICC_DATA" />
```
or we could have defined a set of fields to protect in a **single jPOS environment attribute** and then say

```xml
<property name="protect-ISOMsg" value="${protect.isofields}" />
```

The test cases in `ProtectDebugInfoTest` have also been expanded to consider these cases.

### DIFF
--- a/jpos/src/main/java/org/jpos/transaction/ProtectDebugInfo.java
+++ b/jpos/src/main/java/org/jpos/transaction/ProtectDebugInfo.java
@@ -41,14 +41,17 @@ import java.util.Arrays;
  *        &lt;property name="protect-entry" value="PAN, EXP, REQUEST_ICC_DATA" /&gt;
  *        &lt;property name="wipe-entry"    value="EXPDATE" /&gt;
  *
+ *        &lt;-- if the protected ctx entry is an ISOMsg --&gt;
  *        &lt;property name="protect-ISOMsg" value="2" /&gt;
  *        &lt;property name="protect-ISOMsg" value="35, 45" /&gt;
  *        &lt;property name="wipe-ISOMsg"    value="52, 55" /&gt;
  *
+ *        &lt;-- if the protected ctx entry is a TLVList --&gt;
  *        &lt;property name="wipe-TLVList" value="0x56" /&gt;
  *        &lt;property name="wipe-TLVList" value="0x57" /&gt;
  *        &lt;property name="wipe-TLVList" value="0x5a, 0x5f20" /&gt;
  *
+ *        &lt;-- if the protected ctx entry is a FSDMsg --&gt;
  *        &lt;property name="protect-FSDMsg" value="account-number" /&gt;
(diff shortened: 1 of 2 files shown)


ar-agt:
Thanks, this looks good functionally and the focused tests pass locally:

```bash
./gradlew :jpos:test --tests org.jpos.transaction.ProtectDebugInfoTest
```

One small thing before merge: the updated sample Javadoc now includes:

```xml
<property name="wipe-FSDMsg" value="secret-key" />
```

but `setConfiguration()` still does not read `wipe-FSDMsg`, and the FSD wipe method appears to remain unused. That makes the sample misleading unless this PR also wires `wipe-FSDMsg` support.

I'd suggest either:

1. remove that `wipe-FSDMsg` sample line, or
2. add actual `wipe-FSDMsg` support plus a small test.

Minor wording nit in the Javadoc: “Configuration properties admit...” would read better as “Configuration properties accept...”, and “given in occurrences” as “given as multiple occurrences.”

Other than that, the change looks backward-compatible and useful.

ar-agt:
Thanks for the follow-up. I reviewed the new commit and the `wipe-FSDMsg` wiring now matches the sample configuration. GitHub CI is passing, and the focused local test class still passes:

```bash
./gradlew :jpos:test --tests org.jpos.transaction.ProtectDebugInfoTest
```

One small Javadoc issue remains: the XML comments in the sample are malformed. These lines:

```xml
<-- if the protected ctx entry is an ISOMsg -->
<-- if the protected ctx entry is a TLVList -->
<-- if the protected ctx entry is a FSDMsg -->
```

should be:

```xml
<!-- if the protected ctx entry is an ISOMsg -->
<!-- if the protected ctx entry is a TLVList -->
<!-- if the protected ctx entry is a FSDMsg -->
```

Minor wording nit while there: “Configuration properties admit...” would read better as “Configuration properties accept...”, and “given in occurrences” as “given as multiple occurrences.”

Other than that, looks good to me.

barspi:
Good catch @ar-agt   😄 

Fixed that.

PR #717, opened 2026-05-11
Fix ConcurrentModificationException in ISO components

chhil:
Add thread-safe dump() operations to ISOMsg, TLVList, FSDMsg, ISODatasetField, SecureKeyBlock, SecureKeySpec, CryptographicServiceMessage, and SimpleMsg.

Fixes race conditions where dump() iterates collections (Map.entrySet(), ArrayList) while other threads mutate them via set()/append()/addField().

Pattern applied:
- dump() operations: create snapshot under lock, iterate snapshot
- mutate operations: atomic synchronized blocks

Includes concurrent tests demonstrating the fix with 2000 iterations per thread, 4 threads, and CountDownLatch coordination.

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/ISOMsg.java
+++ b/jpos/src/main/java/org/jpos/iso/ISOMsg.java
@@ -39,10 +39,16 @@ import java.util.*;
 public class ISOMsg extends ISOComponent
     implements Cloneable, Loggeable, Externalizable
 {
-    /** Map of field number to field value. */
+    /** Map of field number to field value. All mutations and iterations must synchronize on this object. */
     protected Map<Integer,Object> fields;
-    /** Highest field number currently set in this message. */
-    protected int maxField;
+    /**
+     * Highest field number currently set in this message.
+     * <p>
+     * Declared {@code volatile} so that any thread reading it (even indirectly via
+     * {@link #recalcMaxField()}) sees the latest value written under {@code synchronized(fields)}.
+     * This prevents stale reads when concurrent threads modify {@code fields}.
+     */
+    protected volatile int maxField;
     /** The packager used to pack/unpack this message. */
     protected ISOPackager packager;
     /** Dirty flags for tracking state changes. */
@@ -190,13 +196,23 @@ public class ISOMsg extends ISOComponent
     }
     /**
      * Returns the highest field number present in this message.
+     * <p>
+     * Synchronizes on {@code fields} to ensure a consistent view of both the
+     * {@code maxFieldDirty} flag and the {@code maxField} value. If the dirty
+     * flag is set, {@link #recalcMaxField()} walks the map keys to recompute
+     * the maximum. This method must use the same lock as {@link #set(ISOComponent)},
+     * {@link #unset(int)}, and {@link #recalcBitMap()} to prevent concurrent
+     * modification and stale reads of {@code maxField}.
+     *
      * @return the max field number
      */
     @Override
     public int getMaxField() {
-        if (maxFieldDirty)
-            recalcMaxField();
-        return maxField;
+        synchronized (fields) {
+            if (maxFieldDirty)
+                recalcMaxField();
+            return maxField;
+        }
     }
 
     /**
@@ -210,6 +226,12 @@ public class ISOMsg extends ISOComponent
             return new TreeSet<>(fields.keySet());
         }
     }
+    /**
+     * Recomputes {@link #maxField} by scanning all keys in the fields map.
+     * <p>
+     * Must only be called from within a {@code synchronized(fields)} block —
+     * callers include {@link #getMaxField()} and {@link #recalcBitMap()}.
+     */
     private void recalcMaxField() {
         maxField = 0;
         for (Object obj : fields.keySet()) {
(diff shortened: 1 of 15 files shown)


jpos/src/main/java/org/jpos/iso/ISOMsg.java
```diff
@@ -232,11 +244,13 @@ public ISOPackager getPackager () {
...
-            if (i > maxField)
-                maxField = i;
-            dirty = true;
+            synchronized (fields) {
+                Integer i = (Integer) c.getKey();
+                fields.put (i, c);
+                if (i > maxField)
+                    maxField = i;
+                dirty = true;
+            }
```
alcarraz, line 286: I know this woun't make much of a difference, but why not putting the `synchronized` block just arround fields.put?

```suggestion
            Integer i = (Integer) c.getKey();
            synchronized (fields) {
                fields.put (i, c);5
            }
            if (i > maxField)
                maxField = i;
            dirty = true;
```

alcarraz (commented):
An alternative to this would be to just state that `dump` is not thread-safe.  My only concern is about the extra pressure we may be putting on the GC.

I don't have a strong opinion on this, and I don't know if this would have an appreciable impact; I'm only mentioning it in case it wasn't considered. This only affects the `dump` changes.

I think the `synchronized` blocks are needed for `set` to be thread-safe. I'm not sure if we need that too, or how much the monitor locking would impact performance, as that method is heavily used.

jpos/src/main/java/org/jpos/iso/ISOMsg.java
chhil: > An alternative to this would be to just state that dump is not thread-safe. 

We see CME's as warning in the log. At least for jpos components it should work, but user ones is not jpos responsibility.

>  but why not putting the synchronized block just arround fields.put?

My test were failing. I distinctly recall having to push the dirty flag inside too.

maxField must stay inside the synchronized block. It has no volatile keyword and no synchronization on read (only via recalcBitMap() → getMaxField()). If maxField = i moved outside the lock, another thread could enter getFieldNumbers() (which does acquire the lock) and call recalcBitMap(), which iterates fields.keySet() — but sees a stale maxField. This would cause the bitmap to be recalculated with an incorrect range, potentially missing fields.

alcarraz: > > but why not putting the synchronized block just arround fields.put?
> 
> My test were failing. I distinctly recall having to push the dirty flag inside too.
> 
> maxField must stay inside the synchronized block. It has no volatile keyword and no synchronization on read (only via recalcBitMap() → getMaxField()). If maxField = i moved outside the lock, another thread could enter getFieldNumbers() (which does acquire the lock) and call recalcBitMap(), which iterates fields.keySet() — but sees a stale maxField. This would cause the bitmap to be recalculated with an incorrect range, potentially missing fields.

Yes, I was thinking only about the CME when I wrote that. Since I didn't see any other place where `synchronized(fields)` was accessing the dirty flags. Shouldn't `recalcBitMap` also be synchronized on `fields` then? And `unset`, `getMaxField`?

ar-agt:
I think we need to be careful with the framing here. `ISOMsg` itself is not currently designed to be thread-safe; in normal jPOS usage, thread isolation is provided at a higher level by the transaction-manager pattern. So I'm reluctant to add partial synchronization to `ISOMsg` and related classes, because it may create a stronger "thread-safe" impression than the codebase actually guarantees, which could lead users into unsafe usage patterns.

I also found at least one concrete issue in the current changes: the new `ISODatasetField` test fails locally. The implementation snapshots `datasets` under `synchronized (datasets)`, but `addDataset`, `removeDataset`, and `setValue` still mutate the same `ArrayList` without that lock, so the synchronization does not reliably protect the collection.

More broadly, several changes appear to protect only the dump-side iteration while leaving other access/mutation paths unsynchronized or using different locking scopes. That makes the fix incomplete if the goal is true thread safety.

Given that, I don't think we should merge this as-is. The safer direction is probably to document/flag that these objects are not thread-safe and should not be mutated while being dumped, rather than introducing partial synchronization that suggests a guarantee we do not actually provide.

This is still a useful report and helped clarify an important boundary in the API.

chhil:
> I also found at least one concrete issue in the current changes: the new `ISODatasetField` test fails locally. The implementation snapshots `datasets` under `synchronized (datasets)`, but `addDataset`, `removeDataset`, and `setValue` still mutate the same `ArrayList` without that lock, so the synchronization does not reliably protect the collection.
> 


Fixed in 9f6f891f3822

ar-agt:
I reviewed the follow-up commit `9f6f891f`. It does address the specific issue I called out in `ISODatasetField`: `addDataset`, `removeDataset`, `setValue`, and lookup paths now use the same `datasets` monitor. The ISOMsg correction also addresses the lock-scope concern raised in review: `pack`, `unpack`, `recalcBitMap`, `unset`, and `getMaxField` now use `synchronized(fields)`.

I also ran the targeted concurrent tests locally:

```sh
./gradlew :jpos:test --tests org.jpos.iso.ISOMsgConcurrentTest --tests org.jpos.iso.ISOMsgPackConcurrentTest
```

They pass under the PR's `.sdkmanrc` (`java=26-amzn`, `gradle=9.5.0`).

That said, my broader concern remains: this still should not be framed as making these components thread-safe. Several dump implementations snapshot only keys, then read values after releasing the lock. For example:

- `ISOMsg.dump()` gets field numbers under lock, then calls `fields.get(i)` outside the lock.
- `FSDMsg.dump()` snapshots field names, then calls `fields.get(f)` outside the lock.
- `CryptographicServiceMessage.dump()` snapshots tags, then calls `getFieldContent(tag)` outside the lock.

That avoids the key-iterator `ConcurrentModificationException`, but it is not a coherent locked snapshot. `CryptographicServiceMessage` is also still internally inconsistent: `addField` is synchronized, but `removeField`, `removeFields`, `getFieldContent`, and `toString` are not, and `toString` still iterates `fields.keySet()` directly.

So I think the correction is real progress, but I would either narrow the stated goal to best-effort CME avoidance during dump, or make the dump snapshots include the values/components under the same lock and make each touched class internally consistent. I would avoid merging this as "thread-safe ISO components" because the implementation still does not provide that guarantee.

PR #725, opened 2026-06-10
Add optional channel connection logging

ar-agt:
## Summary
- add `log-connections` to `BaseChannel`, defaulting to `true`, so legacy successful connect and normal disconnect logging remains enabled unless explicitly disabled
- allow deployments to opt out of routine channel lifecycle logs with `<property name="log-connections" value="false" />` on the `<channel>` element
- keep failed connection attempts, protocol errors, and non-routine I/O failures logged
- add Micrometer counters for total channel/server connection lifecycle events alongside the existing active-connection gauges
- update BaseChannel tests for the legacy default and explicit opt-out behavior

## Compatibility note
The default remains compatible with existing deployments: routine successful connection and normal disconnection events are logged by default. Deployments that want the quieter behavior can set `log-connections=false` on the channel configuration.

## Verification
- `JAVA_HOME=$HOME/.sdkman/candidates/java/26.0.1-amzn ./gradlew :jpos:test --tests org.jpos.iso.BaseChannelTest --tests org.jpos.q2.iso.ChannelAdaptorTest --tests org.jpos.q2.iso.QServerTest`
- `git diff --check`

### DIFF
--- a/jpos/src/main/java/org/jpos/q2/iso/ChannelAdaptor.java
+++ b/jpos/src/main/java/org/jpos/q2/iso/ChannelAdaptor.java
@@ -18,6 +18,7 @@
 
 package org.jpos.q2.iso;
 
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Gauge;
 import io.micrometer.core.instrument.Tags;
 import io.micrometer.core.instrument.binder.BaseUnits;
@@ -84,6 +85,8 @@ public class ChannelAdaptor
     private ScheduledExecutorService scheduledExecutor;
 
     private Gauge connectionsGauge;
+    private Counter connectsCounter;
+    private Counter disconnectsCounter;
 
     @Config("soft-stop") private long softStop;
 
@@ -524,10 +527,13 @@ public class ChannelAdaptor
             } catch (IOException ignored) {
                 // channel.connect already logs - no need for more warnings
             }
-            if (!channel.isConnected ())
+            if (!channel.isConnected ()) {
                 ISOUtil.sleep (delay);
-            else
+            } else {
                 connects++;
+                if (connectsCounter != null)
+                    connectsCounter.increment();
+            }
         }
         if (running() && sp.rdp (ready) == null)
             sp.out (ready, new Date());
@@ -541,6 +547,8 @@ public class ChannelAdaptor
             try {
                 SpaceUtil.wipe(sp, ready);
                 channel.disconnect();
+                if (disconnectsCounter != null)
+                    disconnectsCounter.increment();
             } catch (Exception e) {
                 getLog().warn("disconnect", e);
             }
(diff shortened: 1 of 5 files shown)


jpos/src/main/java/org/jpos/q2/iso/QServer.java
```diff
@@ -400,6 +403,14 @@ private void initMeters() {
...
               server::getActiveConnections
             );
+        acceptsCounter = MeterFactory.counter(registry, MeterInfo.ISOSERVER_ACCEPTS, tags);
+        disconnectsCounter = MeterFactory.counter(registry, MeterInfo.ISOSERVER_DISCONNECTS, tags);
+        server.addServerEventListener(event -> {
+            if (event instanceof ISOServerAcceptEvent && acceptsCounter != null)
+                acceptsCounter.increment();
+            else if (event instanceof ISOServerClientDisconnectEvent && disconnectsCounter != null)
+                disconnectsCounter.increment();
+        });
```
chhil: Should this litener be removed on stopService?

jpos/src/main/java/org/jpos/q2/iso/ChannelAdaptor.java
```diff
@@ -541,6 +547,8 @@ protected void disconnect () {
             try {
                 SpaceUtil.wipe(sp, ready);
                 channel.disconnect();
+                if (disconnectsCounter != null)
+                    disconnectsCounter.increment();
             } catch (Exception e) {
                 getLog().warn("disconnect", e);
             }
```
chhil, line 555: If `channel.disconnect()` throws an exception the `disconnectCounter` wont increment.

ar-agt:
Thanks for catching both of these.

You were right on both points, and I've pushed follow-up fixes on this PR branch:

- `27a4701a0` `fix: remove qserver metrics listener on stop`
  This stores the `QServer` metrics listener and unregisters it during teardown instead of leaving an anonymous listener attached.

- `8b8eff1cc` `fix: count channel disconnects after cleanup errors`
  This moves the `ChannelAdaptor` disconnect counter update into a `finally` path guarded by `!channel.isConnected()`, so cleanup exceptions do not suppress a real disconnect count.

Heads-up: the second fix was a small follow-up after addressing the first comment, so both review points should now be covered by the branch head.

ar-agt:
Configuration note: routine channel connection logging remains enabled by default. To turn it off for a specific channel, add the `log-connections` property with value `false` to that channel configuration:

```xml
<channel class="org.jpos.iso.channel.ASCIIChannel" packager="org.jpos.iso.packager.GenericPackager">
    <property name="log-connections" value="false" />
</channel>
```

For programmatic configuration, call:

```java
channel.setLogConnections(false);
```

This is a per-channel opt-out; omitting the property keeps the legacy/default behavior (`true`).

PR #736, opened 2026-07-14
After 50 million requests in load testing observed out of memory

keshav120:
i was load testing with huge transaction and found the performance degraded after 2 days of running load, the memory is not reclaimed and kubernatics pod filled up slowly and after 50M or 100 million i realised pod was going out of memory. 
After multiple iteration of running load, i found through heap dump / hprof file after using memory analyser. It was Instant.now objects in millions and culprit.

I created a new class by extending TransactionManager.java and override setThreadName and remove the LocalDateTime and Instant.now altogether and since then problem got resolved.

since this is library so i fixed in TransactionManager.setThreadName method directly.

### DIFF
--- a/jpos/src/main/java/org/jpos/transaction/TransactionManager.java
+++ b/jpos/src/main/java/org/jpos/transaction/TransactionManager.java
@@ -96,6 +96,7 @@ public class TransactionManager
     public static final long    MAX_PARTICIPANTS = 1000;  // loop prevention
     /** Default maximum time, in milliseconds, that a session waits for input. */
     public static final long    MAX_WAIT = 15000L;
+    private static final ZoneId LOCAL_ZONE_ID = ZoneId.systemDefault();
     /** Configured group-name to participant-list mapping. */
     protected Map<String,List<TransactionParticipant>> groups;
     private Set<Destroyable> destroyables = new HashSet<>();
@@ -1329,10 +1330,12 @@ public class TransactionManager
         }
     }
     private void setThreadName (long id, String method, TransactionParticipant p) {
-        Thread.currentThread().setName(
-            String.format("%s:%d %s %s %s", getName(), id, method, p.getClass().getName(),
-                LocalDateTime.ofInstant(Instant.now(), ZoneId.systemDefault()))
-        );
+        StringBuilder sb = new StringBuilder(64);
+        sb.append(getName()).append(':').append(id).append(' ')
+          .append(method).append(' ')
+          .append(p.getClass().getName()).append(' ')
+          .append(LocalDateTime.now(LOCAL_ZONE_ID));
+        Thread.currentThread().setName(sb.toString());
     }
     private void setThreadLocal (long id, Serializable context) {
         tlId.set(id);


ar:
Thanks for investigating this and for sharing the load-testing results.

 There are a couple of points we need to clarify before merging.
 
 `LocalDateTime.now(LOCAL_ZONE_ID)` still obtains the current time through
`Clock.instant()`, so the proposed change does not actually eliminate Instant
creation. It does avoid the overhead of `String.format`, which may improve
allocation rates, but that is a different issue.

Also, the temporary Instant created by the existing method is not retained
after the thread-name string is built. If the heap dump contains millions of
live Instant objects, could you please share the relevant dominator or path-
to-GC-root information? That would help confirm where those objects are being
retained.

If the timestamp in the thread name is the source of excessive allocation,
perhaps the simplest fix is to remove it entirely. It would also be helpful to
include before-and-after measurements or a small reproducible load test
demonstrating the improvement.

One additional detail: caching `ZoneId.systemDefault()` changes the behavior if
the JVM’s default timezone is updated while running, so we should avoid that
unless it is intentional.

Thanks again—there may well be a useful optimization here, but we need a
little more evidence to ensure we are fixing the actual cause of the memory
growth.

chhil:
I put my LLM to work for a unit test and jmh test.

[TransactionManagerThreadNameTest.java](https://github.com/user-attachments/files/30112698/TransactionManagerThreadNameTest.java)
[TransactionManagerSetThreadNameBenchmark.java](https://github.com/user-attachments/files/30112704/TransactionManagerSetThreadNameBenchmark.java)


### PR #736 Analysis: TransactionManager.setThreadName Optimization

**PR:** https://github.com/jpos/jPOS/pull/736
**File:** `jpos/src/main/java/org/jpos/transaction/TransactionManager.java` (+7/−4)
**Test file:** `jpos/src/test/java/org/jpos/transaction/TransactionManagerThreadNameTest.java`
**Benchmark:** `jpos/src/test/java/org/jpos/transaction/TransactionManagerSetThreadNameBenchmark.java`

---

## The Claim

> "After 50M requests in load testing observed out of memory due to `Instant.now` ... found through heap dump / hprof file after using memory analyser. It was `Instant.now` objects in millions and culprit."

The fix replaces:

```java
// Original (line 1332-1335)
Thread.currentThread().setName(
    String.format("%s:%d %s %s %s", getName(), id, method, p.getClass().getName(),
        LocalDateTime.ofInstant(Instant.now(), ZoneId.systemDefault()))
);
```

with:

```java
// PR #736
private static final ZoneId LOCAL_ZONE_ID = ZoneId.systemDefault();
// ...
StringBuilder sb = new StringBuilder(64);
sb.append(getName()).append(':').append(id).append(' ')
  .append(method).append(' ')
  .append(p.getClass().getName()).append(' ')
  .append(LocalDateTime.now(LOCAL_ZONE_ID));
Thread.currentThread().setName(sb.toString());
```

---

## JMH Benchmark Methodology

- **JMH version:** 1.37
- **JVM:** OpenJDK 25 Corretto (25.0.3+9-LTS)
- **Forks:** 2 (1 warmup fork, then 2 measurement forks)
- **Warmup:** 3 iterations, 2 seconds each
- **Measurement:** 5 iterations, 3 seconds each
- **Mode:** Throughput (`ops/s`) and Average Time (`us/op`)
- **Blackhole:** Compiler mode
- **Run with:** `./gradlew :jpos:jmhBenchmark`

Six benchmarks isolate the source of every improvement:

| Benchmark | What it measures |
|---|---|
| `dateTimeViaInstantNow` | `LocalDateTime.ofInstant(Instant.now(), zone)` |
| `dateTimeViaNow` | `LocalDateTime.now(zone)` |
| `formatWithoutDate` | `String.format(...)` without any date component |
| `stringBuilderWithoutDate` | `StringBuilder` without any date component |
| `threadNameOld` | Full old approach: `String.format` + `Instant.now()` + `ofInstant()` |
| `threadNameNew` | New approach (PR) with `StringBuilder(64)` |
| `threadNameNewCapacity160` | New approach with `StringBuilder(160)` (no resize) |

---

## Results

### Date/time only: both are virtually identical

| Benchmark | Throughput | Avg Time |
|---|---|---|
| `LocalDateTime.ofInstant(Instant.now(), zone)` | 13.34 ± 4.03 ops/us | **0.065 ± 0.008 us/op** |
| `LocalDateTime.now(zone)` | 16.65 ± 2.51 ops/us | **0.061 ± 0.004 us/op** |

`LocalDateTime.now(ZoneId)` calls `Clock.system(zone).instant()` internally.
Both code paths allocate an `Instant` object at the same rate.
The ~25% throughput difference is within noise range given the wide confidence intervals.
In absolute latency terms, the difference is **0.004 microseconds** — negligible.
**The `Instant.now()` removal is not the source of the improvement.**

### Isolating String overhead: this is the real savings

| Benchmark | Throughput | Ratio |
|---|---|---|
| `String.format(...)` without date | 3,410,016 ± 1,011,557 ops/s | **1x** |
| `StringBuilder` without date | 40,629,724 ± 11,710,989 ops/s | **11.9x** |

`StringBuilder` is **~12x faster** than `String.format` even before considering the date/time allocation. `String.format` allocates a `Formatter` object, varargs `Object[]`, format string parsing machinery, and an internal `StringBuilder`.

### Full thread name: 2.25–2.59x improvement

| Benchmark | Throughput | Ratio vs old |
|---|---|---|
| Old: `String.format` + `Instant.now()` + `ofInstant()` | 2,353,459 ± 362,777 ops/s | **1x** |
| New (capacity 64, PR #736): `StringBuilder` + `LocalDateTime.now()` | 5,292,857 ± 2,553,973 ops/s | **2.25x** |
| New (capacity 160, no resize): `StringBuilder` + `LocalDateTime.now()` | 6,088,500 ± 2,219,483 ops/s | **2.59x** |

The improvement is **2.25x** with the PR's `StringBuilder(64)` and **2.59x** with `StringBuilder(160)`.

**The savings come from replacing `String.format` with `StringBuilder`, not from removing `Instant.now()`.**

---

## Secondary Issue: StringBuilder Capacity

PR #736 uses `new StringBuilder(64)`. Measured thread name lengths:

| Scenario | Length |
|---|---|
| Shortest observed | 72 chars |
| Typical (5-participant TM, `prepare`) | ~85 chars |
| Worst-case (long name, `prepareForAbort`, nanosecond timestamp) | **130 chars** |

**Every call triggers at least one resize** (64 → 130, doubling internal array and copying).
The resize overhead is measurable: capacity 160 gives **15% more throughput** than capacity 64.

**Recommended:** use `new StringBuilder(160)` to avoid resizing on all realistic inputs.

---

## Hot Path Impact

`setThreadName` is called **once per participant per phase** (prepare, then commit or abort):

- 5-participant transaction: **~5–10 calls per transaction**
- 100M transactions: **500M–1B calls**

| Metric | Old approach | New (cap 64) | New (cap 160) |
|---|---|---|---|
| Throughput | 2.35M ops/s | 5.29M ops/s | 6.09M ops/s |
| Speedup vs old | 1x | **2.25x** | **2.59x** |
| Allocation saved per call | — | ~625 bytes | ~655 bytes |

The performance improvement is real and meaningful. It comes from `String.format` → `StringBuilder`, not from `Instant.now()` → `LocalDateTime.now()`.

---

## Test Coverage

**File:** `TransactionManagerThreadNameTest.java` (7 tests, all passing)

| Test | What it proves |
|---|---|
| `oldApproachStillAllocatesInstant` | Both approaches allocate identically — Instant is NOT eliminated |
| `newApproachReducesAllocation` | Overall allocation is reduced (via ThreadMXBean) |
| `threadNameContainsAllParts` | Thread name format is correct (name, id, method, participant, timestamp) |
| `threadNameMatchesOldFormat` | Prefix structure before timestamp is identical |
| `threadNameCapacityShouldBeAdequate` | 64 is too small, 160 fits all cases |
| `multipleParticipantsGetDistinctNames` | Each participant produces a unique name |
| `setThreadNameSetsCorrectName` | Integration test via reflection, restores original thread name |

---

## Conclusions

1. **The fix is directionally correct** — the `StringBuilder` over `String.format` change gives a real 2.25–2.59x throughput improvement on the hot path.

2. **The claimed root cause is wrong** — `Instant.now()` was not the OOM culprit. Both old and new code allocate `Instant` objects at the same rate. The savings come entirely from removing `String.format`.

3. **`StringBuilder(64)` is too small** — the buffer always resizes on realistic thread names. Use `StringBuilder(160)`.

4. **The fix should be reframed** — the PR description should state: "Replaced String.format with StringBuilder in setThreadName hot path to reduce allocation rate by ~60% (2.5x throughput improvement)."

5. **Consider dropping the timestamp entirely** — the author's subclass proof-of-concept removed the timestamp, which would eliminate both the `Instant` and `LocalDateTime` allocations on top of the `StringBuilder` savings. This depends on whether the timestamp in the thread name is needed for debugging.

---

## Running the Tests & Benchmarks

```sh
# Unit tests
./gradlew :jpos:test --tests "org.jpos.transaction.TransactionManagerThreadNameTest"

# JMH benchmarks
./gradlew :jpos:jmhBenchmark
```

--- 

For `jmh` the build and lib files needed to be updated

build.gradle 
```
    testImplementation libs.jmhCore
    testAnnotationProcessor libs.jmhAnnotationProcessor
```
libs.version.toml

```
jmh = '1.37'


jmhCore = { module = "org.openjdk.jmh:jmh-core", version.ref = "jmh" }
jmhAnnotationProcessor = { module = "org.openjdk.jmh:jmh-generator-annprocess", version.ref = "jmh" }
```

ar:
@keshav120 thanks again for the report, and @chhil for taking the time to run the numbers—the benchmark
  actually settles the key point: both the old and the new code allocate an Instant (or equivalent) per
  call, so the premise that `Instant.now()` was the culprit doesn't hold, as your own analysis concludes.

  Stepping back, though, I think the discussion is on the wrong axis. An OOM that develops over days is a
  retention problem, not an allocation-rate problem. Everything `setThreadName` creates—the `Instant`, the
  `LocalDateTime`, the resulting `String`—becomes unreachable microseconds after the call returns, and any
  collector reclaims that trivially. You can't fill a heap with objects that die young, no matter how many
  millions you allocate, and making them 2.5x cheaper to allocate can't fix an OOM. Also note that heap
  dumps include unreachable objects unless captured live-only, so "millions of Instants in the `hprof`" on a
  busy system is expected and isn't by itself evidence of a leak. These are session threads processing
  transactions—even at tens of thousands of TPS we're talking modest allocation rates on this path; a
  microbenchmark of it (or tuning `StringBuilder` capacity) is measuring something that was never the problem.

  The genuinely interesting datapoint is that your production workaround removed the timestamp from the
  thread name, and that resolved the problem. The timestamp is what makes every thread name unique. If
  something in your stack retains state keyed by thread name, unbounded name cardinality produces exactly
  your symptom: slow growth over days, proportional to transaction count, fixed by removing the timestamp—and not fixed by this PR, which keeps it. We've verified jPOS's own Micrometer instrumentation only uses
  bounded tags (TM name, participant, phase, realm) and never touches thread names, so the usual suspects
  would be an APM/observability agent, a log pipeline, or application code mapping by thread name.

  To make progress we still need what was requested earlier: the dominator tree / path-to-GC-roots from a
  live-only heap dump. That will show in one glance what retains those objects and whether thread names are
  involved at all. You can get it with `jmap -dump:live,format=b,file=/tmp/live.hprof <pid>`.

  I'm closing this PR because, as it stands, it wouldn't fix the reported problem, and its title would leave
  a misleading trail in the history. Two things remain on the table:

  - If the dominator tree confirms retention keyed by thread name, the right fix on our side is dropping the
  timestamp from setThreadName (or making it optional)—thread dumps carry their own timestamps anyway.
  That's a small, honestly-titled PR we'd happily take. Please open an issue with the heap-dump evidence and
  we'll take it from there.
  - If we ever touch that line for performance, plain string concatenation is the right replacement for
  String.format on modern JDKs (it compiles to `StringConcatFactory` and performs on par with a hand-tuned
  `StringBuilder`)—no capacity tuning needed for a debugging aid. And caching `ZoneId.systemDefault()` changes
  behavior if the JVM's default timezone changes at runtime, so we'd avoid that regardless.

  Thanks to both of you—the load-testing effort and the benchmarks are genuinely appreciated, even where
  we disagree with the conclusions.

keshav120:
hello @ar and @chhil thanks for the effort by putting on comment.

i m attaching hprof dump screenshot and there i got the assumption idea of this issue. 
IMG20260718080942

Here was my fix .

IMG20260718082646

i had run 4 iteration of load test and every time i observed after 50 or 80  million of transaction, the grafana memory was filling slowly and not reclaimed after gc and after 2-3 days the jvm crashed. 

i fixed by entirely removing the timestamp and this proves my escape hatch.

- The fix i suggested may not be correct but the issue is real.

PR #750, opened 2026-08-11
docs: document Switch constructor

ar-agt:
Moves the existing selector-construction Javadoc from the configuration field to the Switch constructor, making the Javadoc task warning-free.

### DIFF
--- a/jpos/src/main/java/org/jpos/transaction/participant/Switch.java
+++ b/jpos/src/main/java/org/jpos/transaction/participant/Switch.java
@@ -40,12 +40,12 @@ public class Switch implements Configurable, GroupSelector {
     private static final String PREFIX_MODE = "prefix";
     private static final Set<String> SELECTOR_PROPERTIES = Set.of("mode", "txnname", "unknown");
 
-    /** Creates the selector; configuration is supplied via {@link #setConfiguration(Configuration)}. */
     private Configuration cfg;
     private String txnNameEntry;
     private boolean prefixMode;
     private List<String> sortedRoutingKeys = List.of();   // kept sorted from longest to shortest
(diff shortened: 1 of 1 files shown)


jpos/src/main/java/org/jpos/transaction/participant/Switch.java
```diff
@@ -40,12 +40,12 @@ public class Switch implements Configurable, GroupSelector {
     private static final String PREFIX_MODE = "prefix";
     private static final Set<String> SELECTOR_PROPERTIES = Set.of("mode", "txnname", "unknown");
 
-    /** Creates the selector; configuration is supplied via {@link #setConfiguration(Configuration)}. */
     private Configuration cfg;
     private String txnNameEntry;
     private boolean prefixMode;
     private List<String> sortedRoutingKeys = List.of();   // kept sorted from longest to shortest
 
+    /** Creates the selector; configuration is supplied via {@link #setConfiguration(Configuration)}. */
```
alcarraz, line 48: The javadoc is for the `cfg` field. I guess the warning is because the field is Private. I think it could just be removed and document the `setConfiguration` method instead, or just make it a regular Java line comment.

ar: @alcarraz can you correct this PR ?

alcarraz: Actually, I misread the javadoc, sorry for the noise, this is OK, looking up the history this was the original javadoc for the constructor, that was left misplaced by #749.

I will still take the chance to write the javadoc for `setConfiguration` since I'm here.

PR #756, opened 2026-08-30
Fix Prometheus scrape allocations and bound JFR recording

### LINKED ISSUE #755: Idle Q2 instances park at max heap and leak /tmp: oversized Prometheus scrape, non-streaming exposition, unbounded default JFR recording
## Summary

An idle jPOS Control Plane satellite (jcard, 2 Gi limit → 1.33 GB max heap) shows a heap sawtooth: Old Gen climbs ~92 MB/min with no GC activity until it hits the heap ceiling (~1.3 GB), then collapses to ~76 MB. Root cause is the Prometheus scrape path: each 15-second scrape renders a 3.65 MB exposition document and copies it five times (~18.4 MB of `byte[]` ≥ 512 KB per scrape — all humongous for a 1 MB G1 region, all landing directly in Old Gen). Whether a pod shows the sawtooth depends only on whether *other* traffic triggers young GCs (whose eager reclaim hides the humongous garbage), which is why two pods on the same image graph completely differently.

Separately, Q2's always-on default JFR recording has no bounds, so the JFR repository in `/tmp` grows for the life of the JVM (observed: 1.2 GB after 2.2 days on one pod).

All numbers below were measured on demo.jpos.org with JFR (`jdk.ObjectAllocationOutsideTLAB`, exact sizes, not sampled).

## Measurements

Per 15 s scrape of an idle jcard instance:

| Bytes | Allocation site | What |
|---|---|---|
| ~7.6 MB | `Arrays.copyOf` under `PrometheusTextFormatWriter.write` | output buffer doubling 0.5→1→2→4 MB |
| 3.6 MB | `PrometheusMeterRegistry.scrape()` L143 | final copy into the returned `String` |
| 3.6 MB | `PrometheusService` L54 | `response.getBytes()` #1, only for `Content-Length` |
| 3.6 MB | `PrometheusService` L56 | `response.getBytes()` #2, the write |

Payload composition: 3,651 KB total, of which **3,642 KB (99.7 %) is `jpos_tm_op_seconds`** — 345 timers (TM `name × participant × phase`) × ~83 lines each. 80 of those lines are histogram buckets: 74 from `MeterFactory.timer()`'s `publishPercentileHistogram()` plus 6 SLOs *merged* on top by Q2's `TM_OPERATION` `MeterFilter` (`Q2.java` ~1703). On an idle instance all 27,600 buckets are `0.0`.

JFR repository (`Q2.startJFR()`, `default` profile, no `maxSize`/`maxAge`): 258 MB–1.2 GB per pod after 2.2 days, ~120–550 MB/day of ephemeral storage, reclaimed only on JVM restart.

## Work items

1. **Trim the TM op histogram** — in Q2's existing `TM_OPERATION` meter filter, add `.percentilesHistogram(false)` so the 6 SLO buckets replace (not extend) the 74 auto-buckets. Payload drops 3.65 MB → ~0.5 MB (~12 lines/timer). Consider whether the client-side p50/p95 from `publishPercentiles(0.5, 0.95)` in `MeterFactory.timer()` are used anywhere; if not, drop them too.

2. **Stream the scrape** — `PrometheusService`: replace the `String response = registry.scrape(); … getBytes() × 2` pattern with `sendResponseHeaders(200, 0)` (chunked) + `registry.scrape(outputStream)` (available since the current Micrometer, 1.17). Eliminates every humongous allocation in the scrape path regardless of payload size. While there: the existing `getBytes()` uses the platform charset; the status-path handler has the same double-`getBytes()` pattern.

3. **Bound the default JFR recording** — in `Q2.startJFR()`:
   ```java
   recording.setMaxSize(250 * 1024 * 1024);
   recording.setMaxAge(Duration.ofHours(4));
   ```
   Keeps the always-on flight recorder (and `jcmd <pid> JFR.dump` forensics) while capping repo disk; `-Nf/--no-jfr` stays as the opt-out.

## Acceptance

- Idle Q2 with a Prometheus scraper attached holds a flat Old Gen (no ramp to max heap).
- `/metrics` on an idle TM-bearing instance is well under 1 MB.
- JFR repository size stays bounded (≤ maxSize) over multi-day uptime.

---

Notes: item 2 alone fixes the memory sawtooth; item 1 alone fixes the payload/Prometheus cost; they're independent and both worth doing. Item 3 is unrelated to metrics but was found in the same investigation.

ar-agt:
## What Problem This Solves

Idle Q2 instances can accumulate humongous Prometheus scrape allocations in Old Gen, expose an oversized TM operation histogram, and let the always-on JFR repository grow without bounds.

## Why This Change Was Made

- stream Prometheus exposition directly to the HTTP response using chunked transfer encoding instead of building and copying a large `String`
- disable Micrometer's automatic percentile histogram for `jpos.tm.op`, retaining the six explicit SLO buckets
- cap the default JFR recording at 250 MiB and four hours
- encode the small status response once with UTF-8

## User Impact

Prometheus scrapes no longer create repeated full-payload byte arrays, TM-bearing instances expose a substantially smaller metrics document, and Q2's default JFR repository remains bounded during long uptimes.

## Evidence

- `./gradlew :jpos:test --tests org.jpos.q2.Q2Test --no-daemon`
- `./gradlew :jpos:test --no-daemon`
- regression test verifies that TM operation timers expose only the six configured SLO buckets plus `+Inf`

Fixes jpos/jPOS#755

### DIFF
--- a/jpos/src/main/java/org/jpos/metrics/PrometheusService.java
+++ b/jpos/src/main/java/org/jpos/metrics/PrometheusService.java
@@ -27,6 +27,7 @@ import org.jpos.q2.QBeanSupport;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.net.InetSocketAddress;
+import java.nio.charset.StandardCharsets;
 
 /**
  * QBean that exposes the Q2 Prometheus meter registry over an embedded HTTP
@@ -49,20 +50,20 @@ public class PrometheusService extends QBeanSupport {
             final var registry =  getServer().getPrometheusMeterRegistry();
             server = HttpServer.create(new InetSocketAddress(port), 0);
             server.createContext(path, httpExchange -> {
-                String response = registry.scrape();
                 httpExchange.getResponseHeaders().add("Content-Type", "text/plain; version=0.0.4");
-                httpExchange.sendResponseHeaders(200, response.getBytes().length);
+                httpExchange.sendResponseHeaders(200, 0);
                 try (OutputStream os = httpExchange.getResponseBody()) {
-                    os.write(response.getBytes());
+                    registry.scrape(os);
                 }
             });
             if (statusPath != null) {
                 server.createContext(statusPath, httpExchange -> {
-                    String response = getServer().running() ? "running\n" : "stopping\n";
+                    byte[] response = (getServer().running() ? "running\n" : "stopping\n")
+                      .getBytes(StandardCharsets.UTF_8);
                     httpExchange.getResponseHeaders().add("Content-Type", "text/plain");
-                    httpExchange.sendResponseHeaders(200, response.getBytes().length);
+                    httpExchange.sendResponseHeaders(200, response.length);
                     try (OutputStream os = httpExchange.getResponseBody()) {
-                        os.write(response.getBytes());
+                        os.write(response);
                     }
                 });
             }
--- a/jpos/src/main/java/org/jpos/q2/Q2.java
+++ b/jpos/src/main/java/org/jpos/q2/Q2.java
@@ -1678,6 +1678,8 @@ public class Q2 implements FileFilter, Runnable {
         try {
             if (!disableJFR) {
                 recording = new Recording(Configuration.getConfiguration("default"));
+                recording.setMaxSize(250L * 1024 * 1024);
+                recording.setMaxAge(Duration.ofHours(4));
                 recording.start();
             }
         } catch (IOException | ParseException e) {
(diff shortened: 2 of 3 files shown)


ar:
Non-blocking observation on `testTMOperationHistogramUsesOnlyServiceLevelObjectives`: it asserts through the process-global registry. `Q2.meterRegistry` is `Metrics.globalRegistry`, and every `Q2` constructed anywhere in the test JVM runs `registerMicroMeter()`, which `clear()`s that shared registry and stacks one more copy of the `TM_OPERATION` filter plus one more `PrometheusMeterRegistry` member onto it. The test passes today because the timer is registered and scraped inside one method, but it turns order-sensitive the moment another test constructs a `Q2` between registration and assertion (several tests in `Q2Test` do construct extra instances), or registers a conflicting `jpos.tm.op` meter.

Two levels of improvement:

- **Test-local hardening (cheap, could go in this PR):** tag the timer with a unique per-run value (e.g. a UUID) and remove the meter in a `finally`, so nothing leaks into the shared registry for later tests.
- **Actual fix (follow-up material, not this PR):** give each `Q2` instance its own `CompositeMeterRegistry` instead of mutating `Metrics.globalRegistry` from the constructor path. That makes `registerMicroMeter()` idempotent per instance, stops filter/member accumulation when multiple `Q2`s live in one JVM (embedded users, not just tests), and lets tests assert against a fully isolated registry.

ar-agt:
Addressed in 268801881. The regression test now uses a per-run UUID tag and removes its timer from the shared registry in a `finally`, avoiding collisions and meter leakage between tests.

I kept per-`Q2` registry ownership out of this PR as suggested; that broader change needs a separate compatibility and lifecycle discussion.

Verification:

- `./gradlew :jpos:test --tests org.jpos.q2.Q2Test --no-daemon`
- `./gradlew jpos:check --no-daemon`

PR #779, opened 2026-09-09
fix: ISODate.parseISODate no longer corrupts Feb 29th into March 1st

### LINKED ISSUE #778: ISODate.parseISODate's nearest-year inference silently corrupts Feb 29 into March 1st
`ISODate.parseISODate(String d, long currentTime, TimeZone timeZone)`'s year-less branch (`d.length() == 10`, no explicit year) mutates a single shared `Calendar` across all three year candidates:

https://github.com/jpos/jPOS/blob/91c3ac10d1/jpos/src/main/java/org/jpos/iso/ISODate.java#L208-L222

```java
Date thisYear = cal.getTime();
cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)-1);
Date previousYear = cal.getTime();
cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)+2);
Date nextYear = cal.getTime();
if (Math.abs (now.getTime() - previousYear.getTime()) <
    Math.abs (now.getTime() - thisYear.getTime()))
{
    thisYear = previousYear;
} else if (Math.abs (now.getTime() - thisYear.getTime()) >
           Math.abs (now.getTime() - nextYear.getTime()))
{
    thisYear = nextYear;
}
return thisYear;
```

## The bug

`Calendar.set(MONTH, ...)`/`set(DATE, ...)` are lazy: the fields aren't normalized until the next `getTime()`/`get()` call. `GregorianCalendar` is lenient by default, so when `MM/DD = "0229"` and the *real current year* (from `now`, before any of the three candidates are even considered) isn't a leap year, the very first `thisYear = cal.getTime()` call silently rolls Feb 29 over to March 1st, **in place, on the shared `cal` object**.

Every subsequent `cal.set(Calendar.YEAR, ...)` + `getTime()` call for `previousYear`/`nextYear` is computed from that already-corrupted internal state, not from the original `MM=1 (Feb), DD=29` the caller asked for. So even when one of the two other candidate years genuinely is a leap year, where Feb 29 is a real, valid date, the method never recovers it. All three candidates come back March 1st.

## Reproduction

```java
import org.jpos.iso.ISODate;
import java.util.Calendar;
import java.text.SimpleDateFormat;

Calendar now = Calendar.getInstance();
now.set(2025, Calendar.JANUARY, 10, 9, 0, 0);
now.set(Calendar.MILLISECOND, 0);

java.util.Date result = ISODate.parseISODate("0229090000", now.getTimeInMillis());
System.out.println(new SimpleDateFormat("yyyy-MM-dd").format(result));
```

Expected: `2024-02-29`, the nearest Feb 29 to 2025-01-10 is the previous year, 2024, a real leap year.

Actual: `2025-03-01`.

The same corruption happens for the "this year" and "next year" candidates too; which one shows up in the wrong result just depends on which one the (already-broken) distance comparison picks as "nearest."

It also affects the back half of a leap year itself, not just the surrounding non-leap years. `thisYear`'s own candidate is computed first and is genuinely uncorrupted when the real current year is leap, but the distance comparison still prefers whichever candidate is chronologically nearer to `now` overall, and past roughly the midpoint of a leap year, `nextYear` (the following year, never itself leap) is chronologically closer than the already-passed `thisYear` Feb 29, so the corrupted `nextYear` candidate wins over the correct one that was already sitting there:

```java
Calendar now = Calendar.getInstance();
now.set(2024, Calendar.OCTOBER, 15, 12, 0, 0); // 2024 is itself a leap year
now.set(Calendar.MILLISECOND, 0);

java.util.Date result = ISODate.parseISODate("0229120000", now.getTimeInMillis());
System.out.println(new SimpleDateFormat("yyyy-MM-dd").format(result));
```

Expected: `2024-02-29` (the correct, uncorrupted `thisYear` candidate was computed correctly, it's just farther away than the wrong one).

Actual: `2025-03-01`.

## Impact

`parseISODate(String)` (no explicit year) is the standard way to resolve a year-less ISO 8583 date/time field pair: a typical use is DE12 (local transaction time, `HHmmss`) + DE13 (local transaction date, `MMdd`) concatenated into one 10-character string, or DE7 (transmission date/time, `MMddHHmmss`) alone. Any transaction whose local date genuinely is February 29th gets silently stamped with the wrong date, no exception, no error return, nothing to signal that the value is wrong. Downstream systems (settlement, reconciliation, audit) see a plausible but incorrect date, off by up to a year.

This reproduces on every leap day, in every year that isn't itself a leap year (3 years out of 4), plus roughly the second half of the leap year itself: the correct answer briefly exists, uncorrupted, but the nearest-distance comparison discards it anyway.

## Suggested fix

Compute each of the three candidates from an independent `Calendar`/`GregorianCalendar` instance (or via `java.time`, which validates Feb 29 against `Year.isLeap` directly instead of silently rolling over) rather than mutating one shared instance sequentially. That removes the carryover entirely; a non-leap candidate year would then need its own explicit handling (skip it, or treat it as "no valid date this year" rather than silently substituting March 1st) since Feb 29 genuinely doesn't exist there.

Happy to open a PR if a fix along those lines is welcome.

alcarraz:
Fixes #778.

## The Problem
`parseISODate(String, long, TimeZone)`'s year-less branch resolves a `MMDDhhmmss` field into the nearest of three candidate years (previous/current/next) by mutating a single shared `Calendar` in sequence:

```java
Date thisYear = cal.getTime();
cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)-1);
Date previousYear = cal.getTime();
cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)+2);
Date nextYear = cal.getTime();
```

`Calendar` field mutations are lazy: `set()` doesn't normalize anything by itself. Normalization only happens on `getTime()`/`get()`, and when it does, it writes the normalized values back into the `Calendar`'s own field array. So when `MM/DD = "0229"` and the real current year isn't a leap year, the first `thisYear = cal.getTime()` call silently rolls Feb 29th over to March 1st **and mutates `cal` in place**. Every later `cal.set(YEAR, ...)` for `previousYear`/`nextYear` then builds on that already-corrupted `March 1st` state instead of the original `Feb 29th` — so even a genuinely valid leap-year candidate among the three never gets a chance to be computed correctly. All three candidates can come back March 1st, silently, with no error.

## The Fix

Each of the three candidate years is now computed independently, and a genuine date is preferred over one that only exists via rollover.

- `buildDate(cal, year, month, day)` re-asserts `YEAR`/`MONTH`/`DATE` — the only fields a date rollover can touch — before every `getTime()` call, so one candidate's rollover can no longer leak into the next.

  It reuses the existing `cal` instance rather than allocating a fresh `Calendar` per candidate. `parseISODate` sits on a hot path — field 7/12/13 date parsing happens on essentially every ISO-8583 message — and a naive per-candidate fix (a brand-new `Calendar` for each of the three years, `clear()`ed to guarantee no shared state) would add 3-6 extra allocations per call versus the original's single shared instance. Reusing `cal` keeps allocation count unchanged from before the fix, and it's still safe: `HOUR_OF_DAY`/`MINUTE`/`SECOND`/`MILLISECOND` are set once above the branch and a date-only rollover never touches them, so re-asserting only `YEAR`/`MONTH`/`DATE` per candidate is sufficient to erase any state left over from the previous candidate's `getTime()` call.
- `diff(cal, now, candidate, year, month, day)` computes the distance to `now` as before, but adds a large penalty when `day` doesn't actually exist in `month`/`year` (checked via `Calendar.getActualMaximum(DATE)`). This makes a real Feb 29th win the nearest-year comparison over a chronologically-closer-but-rolled-over March 1st, which is what the original bug reports as the expected behavior. The penalty is the same constant for all three candidates, so when none of the three years is a leap year, it cancels out of the comparison and the method falls back to the old rollover-based nearest-year result — the return contract (always a non-null `Date`) is unchanged, since library callers with no null-check shouldn't see new NPEs.

The 3-way candidate selection itself (`if (previousDiff < thisDiff) ... if (thisDiff > nextDiff) ...`) keeps the same shape as the original code, just fed by the corrected `diff` values.

## Testing

Added to `ISODateTest`:
- The two reproduction cases from #778 (a non-leap "current year" with a valid leap year one year away in each direction) — both now resolve to the genuine Feb 29th instead of March 1st.
- A case where none of the three neighboring years is a leap year, confirming the old rollover fallback (March 1st) still applies rather than returning `null` for backward compatibility.

`./gradlew jpos:test` (full suite) passes.

### DIFF
--- a/jpos/src/main/java/org/jpos/iso/ISODate.java
+++ b/jpos/src/main/java/org/jpos/iso/ISODate.java
@@ -203,24 +203,57 @@ public class ISODate {
         if (YY != 0) {
             cal.set (Calendar.YEAR, YY);
             return cal.getTime();
-        } 
+        }
         else {
-            Date thisYear = cal.getTime();
-            cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)-1);
-            Date previousYear = cal.getTime();
-            cal.set (Calendar.YEAR, cal.get (Calendar.YEAR)+2);
-            Date nextYear = cal.getTime();
-            if (Math.abs (now.getTime() - previousYear.getTime()) <
-                Math.abs (now.getTime() - thisYear.getTime())) 
-            {
+            int currentYear = cal.get (Calendar.YEAR);
+            Date previousYear = buildDate (cal, currentYear-1, MM, DD);
+            Date thisYear     = buildDate (cal, currentYear,   MM, DD);
+            Date nextYear     = buildDate (cal, currentYear+1, MM, DD);
+
+            long previousDiff = diff (cal, now, previousYear, currentYear-1, MM, DD);
+            long thisDiff     = diff (cal, now, thisYear,     currentYear,   MM, DD);
+            long nextDiff     = diff (cal, now, nextYear,     currentYear+1, MM, DD);
+
+            if (previousDiff < thisDiff) {
                 thisYear = previousYear;
-            } else if (Math.abs (now.getTime() - thisYear.getTime()) >
-                       Math.abs (now.getTime() - nextYear.getTime()))
-            {
+                thisDiff = previousDiff;
+            }
+            if (thisDiff > nextDiff) {
                 thisYear = nextYear;
             }
             return thisYear;
         }
+    }
+
+    /**
+     * Builds a Date for year/month/day on the given Calendar, re-asserting
+     * YEAR/MONTH/DATE (the only fields a Feb 29th rollover can touch) so
+     * one candidate's rollover can't leak into the next one computed on
+     * the same reused Calendar; HOUR_OF_DAY/MINUTE/SECOND/MILLISECOND are
+     * set once by the caller and untouched by a date-only rollover.
+     */
+    private static Date buildDate (Calendar cal, int year, int month, int day) {
+        cal.set (Calendar.MONTH, month);
+        cal.set (Calendar.DATE, day);
+        cal.set (Calendar.YEAR, year);
+        return cal.getTime();
+    }
+
+    /**
+     * Distance from now to candidate, penalized if day doesn't actually
+     * exist in month/year (e.g. Feb 29th outside a leap year), so a
+     * genuine date is preferred over one that only exists via rollover;
+     * the penalty cancels out of the comparison when every candidate is
+     * equally invalid, falling back to the plain rollover distance.
+     */
+    private static long diff (Calendar cal, Date now, Date candidate, int year, int month, int day) {
+        long diff = Math.abs (now.getTime() - candidate.getTime());
+        cal.set (Calendar.YEAR, year);
+        cal.set (Calendar.MONTH, month);
+        if (day > cal.getActualMaximum(Calendar.DATE)) {
+            diff += Long.MAX_VALUE / 2;   // Big penalty, but low enough to avoid overflow
+        }
+        return diff;
     }
 
     /**
--- a/jpos/src/main/java/org/jpos/iso/ISODate.java
+++ b/jpos/src/main/java/org/jpos/iso/ISODate.java
@@ -21,6 +21,7 @@ package org.jpos.iso;
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.time.Duration;
+import java.time.Year;
 import java.util.*;
 
 /**
@@ -206,13 +207,13 @@ public class ISODate {
         }
         else {
             int currentYear = cal.get (Calendar.YEAR);
-            Date previousYear = buildDate (cal, currentYear-1, MM, DD);
-            Date thisYear     = buildDate (cal, currentYear,   MM, DD);
-            Date nextYear     = buildDate (cal, currentYear+1, MM, DD);
+            Date previousYear = buildDate (cal, currentYear-1, MM, DD, hh, mm, ss);
+            Date thisYear     = buildDate (cal, currentYear,   MM, DD, hh, mm, ss);
+            Date nextYear     = buildDate (cal, currentYear+1, MM, DD, hh, mm, ss);
 
-            long previousDiff = diff (cal, now, previousYear, currentYear-1, MM, DD);
-            long thisDiff     = diff (cal, now, thisYear,     currentYear,   MM, DD);
-            long nextDiff     = diff (cal, now, nextYear,     currentYear+1, MM, DD);
+            long previousDiff = diff (now, previousYear, currentYear-1, MM, DD);
+            long thisDiff     = diff (now, thisYear,     currentYear,   MM, DD);
+            long nextDiff     = diff (now, nextYear,     currentYear+1, MM, DD);
 
             if (previousDiff < thisDiff) {
                 thisYear = previousYear;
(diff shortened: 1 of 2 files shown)


ar-agt (changes requested):
@alcarraz The leap-day cases are fixed, but I reproduced a new DST regression on this head that should be addressed before merging.

`buildDate()` restores only YEAR/MONTH/DATE. Normalizing a candidate in a DST gap also changes HOUR_OF_DAY on the shared Calendar, and that changed hour carries into later candidates.

### Reproduction

```java
ZoneId zone = ZoneId.of("America/New_York");
long now = ZonedDateTime.of(2026, 3, 9, 12, 0, 0, 0, zone)
    .toInstant().toEpochMilli();
Date result = ISODate.parseISODate("0309023000", now, TimeZone.getTimeZone(zone));
System.out.println(result.toInstant().atZone(zone));
```

- Expected and unchanged main (`91c3ac10d`): `2026-03-09T02:30-04:00[America/New_York]`
- This PR (`ef72bc834`): `2026-03-09T03:30-04:00[America/New_York]`

The previous-year candidate, March 9, 2025 at 02:30, falls in a DST gap. Its normalization changes the shared hour to 03:30, contaminating the valid 2026 candidate.

Please capture the reference year before setting the input fields and initialize every candidate from a clean state using all original date/time fields. Reusing one Calendar with `clear()` and complete field initialization is an option; separate instances are another. The validity calculation should likewise not depend on fields left behind by candidate normalization.

Please add this DST regression with an explicit time zone and a full date/time assertion, plus a case where the valid leap day is in the next candidate year. The added leap-day cases currently cover previous/current leap years and the no-leap fallback, not the next-year case.

Verification: all 21 `ISODateTest` tests passed locally using the pinned Java 26.0.2 / Gradle 9.7.1 toolchain. A separate harness compiled against the PR and unchanged main reproduced the different outputs above. I did not rerun the full suite.

alcarraz:
Fixed. `buildDate()` now `clear()`s and fully re-sets every field per candidate (not just `YEAR`/`MONTH`/`DATE`), so a DST-gap normalization can't leak `HOUR_OF_DAY` into the next candidate either. The day-validity check also moved off `Calendar` entirely onto plain `java.time.Year.isLeap` arithmetic, so it can't read leftover state at all.

Also found and fixed a related bug: the validity check indexed a days-in-month array by month without bounds-checking, so a malformed month (`"00"`/`"13"+`) threw `ArrayIndexOutOfBoundsException` instead of falling back to the old lenient rollover like every other malformed input.

Added tests for your DST repro, a next-year leap candidate, and both out-of-range month cases. All pass, `ISODateTest` and the full suite.

alcarraz:
I'd missed the first point of your suggested correction: the reference year was still being read via the same shared `Calendar` that later had the input's `MM`/`DD`/`hh`/`mm`/`ss` written onto it, so a rollover in those input fields (reproducible with a malformed month like `"00"` or `"13"`, not just Feb 29th) could shift the computed "current year" before it was ever read — the same class of ordering hazard as the DST leak, just on a different field.

Rather than just reordering the one read, the last commit splits the shared `Calendar` into one scoped to each branch (explicit-year vs. year-less), so nothing set in one branch can affect a read anywhere else — removes the ordering hazard by construction instead of by convention.

While in there, the three year-less candidates and their comparison now carry plain epoch millis instead of `Date` objects until the winning one is chosen, so it's down to one `Date` instantiation instead of three.

Added regression tests for out-of-range months (too small and too large), confirming the computed reference year now matches the real current year rather than one shifted by the input's own rollover.

PR #780, opened 2026-09-11
Remove redundant isofield with id 96

it240884sii:
Removed redundant ISO field definition for id 96.

### DIFF
--- a/jpos/src/main/resources/packager/cmf-858.xml
+++ b/jpos/src/main/resources/packager/cmf-858.xml
@@ -763,7 +763,7 @@
       name="Reserved for ISO use"
       pad="false"
       class="org.jpos.iso.IFB_LLLLBINARY"/>
-  <isofield id="96"
+  <isofield id="91"
       length="9999"
       name="Reserved for ISO use"
       pad="false"
--- a/jpos/src/main/resources/packager/cmf.xml
+++ b/jpos/src/main/resources/packager/cmf.xml
@@ -763,11 +763,6 @@
       name="Reserved for ISO use"
       pad="false"
       class="org.jpos.iso.IFB_LLLLBINARY"/>
-  <isofield id="96"
-      length="9999"
-      name="Reserved for ISO use"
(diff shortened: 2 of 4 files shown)


ar:
Good catch @it240884sii. I see we have the same problem in cmfv3.xml (copied from this one). That 96 should have been 91. I'll amend your PR and take the opportunity to make the same change in cmfv3.xml shortly. Thank you!

ar-agt:
The duplicate is an original numbering error, rather than a later expansion of field 96. Both entries in `cmf.xml` were introduced together by b2a9bce07a2d9d259e2aa93f9257a31f6afa44c8 (April 1, 2017, “jPOS Common Message Format”). The earlier entry sits between 90 and 92, with 91 absent. The CMF definition reserves 76–92 for binary data up to 9,999 bytes; DE96 is key-management data up to 999 bytes.

`GenericPackager` stores definitions by field ID, so the later DE96 already wins: `IFB_LLLBINARY`, length 999. Deleting the earlier entry preserves that effective DE96 behavior, but leaves DE91 missing. The intended correction is to restore DE91 as `IFB_LLLLBINARY`, length 9999.

The same hole/duplicate exists in `cmfv3.xml` (where both DE96 entries currently describe 999-byte key-management data) and `cmf-858.xml`. At APR's request, I have updated this same PR to correct all three, without changing the effective DE96 definition. Regression coverage checks unique top-level field IDs, maximum-length DE91/DE96 message round trips, their configured maximum lengths, and their BCD length prefixes.

For clarity, LLL and LLLL describe decimal length digits; both binary packagers use two BCD prefix bytes.

Pushed as `c55187fcf` on the existing PR branch (no force push). Local verification with the pinned Java 26.0.2 / Gradle 9.7.1 toolchain: `:jpos:test --tests org.jpos.iso.packager.* :jpos:assemble` passed. 330 tests reported, zero failures/errors, four skipped; includes six new parameterized regression cases. The full repository test suite was not rerun.

